PR #20, opened 2023-07-27
HTML snapshot test for all samples for consistency in the markup

divdavem:


e2e/htmlSnapshot.ts
```diff
@@ -0,0 +1,188 @@
...
+			}
+			return {tagName, childNodes, attributes};
+		};
+		return recFn(element);
+	});
+
+const cleanChildNodes = (childNodes: HTMLNode[]) => {
+	childNodes = [...childNodes];
+	const result: HTMLNode[] = [];
+	for (let i = 0; i < childNodes.length; i++) {
```
quentinderoubaix, line 64: Could this for loop not be rewritten using a while loop ?
```typescript
while (childNodes.length) {
  let newValue = childNodes.shift();
  if (typeof newValue === 'string') {
    const previousValue = result[result.length - 1];
    if (typeof previousValue === 'string' ) {
      result.pop();
      newValue = previousValue + newValue;
    } else if (newValue.tagName === '') {
      childNodes.unshift(...newValue.childNoes);
      continue;
    }
    result.push(newValue); 
  }
}
```

e2e/htmlSnapshot.ts
```diff
@@ -0,0 +1,188 @@
...
+		const recFn = (node: Node) => {
+			if (!isElement(node)) {
+				if (node.nodeType === node.TEXT_NODE) {
+					return node.nodeValue;
+				}
+				// Note: ignore comments
+				return null;
+			}
+			const tagName = node.tagName.toLowerCase();
+			const textarea = isTextArea(node);
```
quentinderoubaix, line 27: minor: could be renamed to ```isNodeArea``` or is ```isNodeTextArea``` to respect naming conventions on boolean variables

e2e/htmlSnapshot.ts
```diff
@@ -0,0 +1,188 @@
...
+			const childNodes: any[] = [];
```
quentinderoubaix: could be typed to ```HTMLNode```

quentinderoubaix (commented):
just a few minor points

PR #23, opened 2023-07-28
test: Record traces and videos on retry only

fbasso:
Just to see what are the performance impact on the test.

Anyway, to would be good to set it this way, as there is no point to record a video for passing tests. The same for the trace, data are recorded during the test even if it's removed if the test passes.


quentinderoubaix (approved):
LGTM !

PR #41, opened 2023-08-01
demo: service worker for offline access

divdavem:


demo/src/service-worker.ts
```diff
@@ -0,0 +1,65 @@
...
+			await self.skipWaiting();
+		})()
+	);
+});
+
+self.addEventListener('activate', (event) => {
+	event.waitUntil(
+		(async () => {
+			await self.clients.claim();
+			for (const key of await caches.keys()) {
```
quentinderoubaix, line 47: If ```caches``` allows it, you could delete all deprecated caches in parallel:
```typescript
await Promise.all((await caches.keys()).filter((key) => key !== CACHE).map((key) => caches.delete(key)));
```

demo/scripts/copySamples.plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+			for (const framework of ['angular', 'react', 'svelte']) {
+				const src = path.join(__dirname, `../../${framework}/dist/demo/`);
+				const dst = `${framework}/samples/`;
+				const files = await glob('**/*', {
+					cwd: src,
+					nodir: true,
+				});
+				for (const file of files) {
+					const name = path.posix.join(dst, file);
+					console.log(name);
```
quentinderoubaix, line 20: You could use instead a logger created using vite ```createLogger``` utility.

demo/src/service-worker.ts
```diff
@@ -0,0 +1,65 @@
...
+	return {withHash, withoutHash};
+};
+
+self.addEventListener('install', (event) => {
+	event.waitUntil(
+		(async () => {
+			const cache = await caches.open(CACHE);
+			const {withHash, withoutHash} = splitAssets(ASSETS);
+			const missingWithHash: string[] = [];
+			for (const url of withHash) {
```
quentinderoubaix, line 29: Maybe could be written as the following:
```typescript
await Promise.all(
	withHash.map((url) =>
		caches.match(url).then((response) => {
			if (response?.ok) {
				cache.put(url, response);
			} else {
				missingWithHash.push(url);
			}
		})
	)
);
```
Not sure of the gain however.

divdavem:
@quentinderoubaix Thank you for your good suggestions. I have added a commit to implement these changes.

quentinderoubaix (approved):
LGTM !

PR #53, opened 2023-08-02
closes #45 generic alert flicker

### LINKED ISSUE #45: Alert dismiss button flickers in React default examples
**Describe the bug**
When navigating to the react examples of the alert widget, the dismiss buttons appear briefly.

**To Reproduce**
Navigate to https://amadeusitgroup.github.io/AgnosUI/latest/react/components/alert/examples.
Refresh or go to the Angular view then back to React. Observe the dismiss buttons flickering.

quentinderoubaix:
Following the recommendations of [this article](https://react.dev/learn/you-might-not-need-an-effect), we do not require a ```useEffect``` when updating the config store and was the root cause of the flicker seen in issue #45.


quentinderoubaix:
@divdavem thx for the review !

PR #67, opened 2023-08-03
build: deterministic sveltekit build

divdavem:
Two builds from the same git commit should produce the same output.

cf https://kit.svelte.dev/docs/configuration#version


quentinderoubaix (approved):
LGTM!

PR #79, opened 2023-08-09
feat: lazy loaded routes for samples

quentinderoubaix:


demo/src/lib/layout/iframe.ts
```diff
@@ -0,0 +1,96 @@
...
+export function createIframeHandler(defaultHeight: number, resize = true, messageType = 'sampleload', spinnerDebounce = 300) {
+	const _iframeLoaded$ = writable(true);
+	const _showSpinner$ = writable(false);
+	const _height$ = writable(0);
+
+	let spinnerTimer: any;
+	let resizeObserver: ResizeObserver | undefined;
+	let heightSubscription: (UnsubscribeFunction & UnsubscribeObject) | undefined;
+
+	const setupObserver = (iframe: HTMLIFrameElement) => {
```
divdavem, line 25: As discussed, this should be extracted in a reactive utility (similar to the intersection utility) that could also be used in the slider.

divdavem (approved):
Thank you for your PR!
Let's merge it to already benefit from the improvements.
We can improve the code of the the iframe utility later (to make it more reactive).

demo/src/lib/layout/iframe.ts
quentinderoubaix: Tracked in #291

PR #96, opened 2023-08-18
Demo improvements (code presented, section headers, readme)

fbasso:
Some changes:

- The code for the samples is now below the example, to avoid that the example watched by the user moves when you toggle the code
- The 'Code' text has been removed to keep the consistency with the other buttons
- h2/h3 font size have been decreased
- A link to the demo site has been added at the first paragraph. Before, the link was only at the bottom of the page, which is (I think) a little bit annoying for a user.
- 
What do you think ?


divdavem (approved):

fbasso:
Thanks @divdavem !

PR #97, opened 2023-08-18
fix: adding back the service worker

divdavem:
It was mistakenly removed in 4c71c4a905c5b52564f59233d05367a0ba91b855 (#82)


ExFlo (approved):

divdavem:
@ExFlo Thank you for your review!

PR #106, opened 2023-08-23
fix(svelte): partly fixes SSR with svelte

divdavem:
Another part of the fix has to be done on Svelte side, cf https://github.com/sveltejs/svelte/issues/9137


fbasso (approved):
LGTM ;)

divdavem:

PR #107, opened 2023-08-23
fix(demo): react custom pagination demo was not consistent with angular and svelte

divdavem:
Changing the value of the second custom pagination example did not change the value of the first one, unlike the angular and svelte examples.


fbasso (approved):
Thanks @divdavem ;)

divdavem:

PR #108, opened 2023-08-24
refactor(page-objects): use getters instead of methods when no parameter is needed

divdavem:


divdavem:

PR #109, opened 2023-08-24
first draft of progressbar

quentinderoubaix:


page-objects/lib/progressbar.po.ts
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export class ProgressbarPO extends BasePO {
+	selectors = structuredClone(progressbarSelectors);
+
+	override getComponentSelector(): string {
+		return this.selectors.rootComponent;
+	}
+
+	locatorOuterBar(): Locator {
```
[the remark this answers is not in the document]
quentinderoubaix: Thanks @fbasso, have corrected.

core/lib/progressbar.ts
```diff
@@ -0,0 +1,191 @@
...
+	 * the minimum value
+	 * @defaultValue 0
+	 */
+	minimum: number;
+
+	/**
+	 * the maximum value
+	 * @defaultValue 100
+	 */
+	maximum: number;
```
fbasso, line 20: Maybe max / min would be a more common name (as Math.min / Math.max for example).

(something to do in the rating, as it does not follow this convention)

core/lib/progressbar.ts
```diff
@@ -0,0 +1,191 @@
...
+			striped$,
+			animated$,
+			showPercentage$,
+			...stateProps
+		},
+		patch,
+	] = writablesForProps(defaultConfig, config, configValidator);
+
+	const maximum$ = computed(() => Math.max(minimum$(), _dirtyMaximum$()));
+	const value$ = computed(() => clamp(_dirtyValue$(), maximum$(), minimum$()));
```
fbasso, line 154: When a value is "rectified" depending on its _dirtyValue and something else, I think it's better to use the `bindableDerived` pattern, as in the rating.

The point is that the dirty value is changed back to its correct value if needed.

core/lib/progressbar.ts
```diff
@@ -0,0 +1,191 @@
...
+ * @returns an ProgressbarWidget
+ */
+export function createProgressbar(config?: PropsConfig<ProgressbarProps>): ProgressbarWidget {
+	const [
+		{
+			minimum$,
+			maximum$: _dirtyMaximum$,
+			// dirty value that needs adjustment:
+			value$: _dirtyValue$,
+			ariaLabel$,
```
fbasso, line 142: No need to create local stores if there are just "proxied" to the return, just get the ones needed in the internal logic.

core/lib/progressbar.ts
quentinderoubaix: Have updated to use the ```bindableDerived``` pattern, thx !

core/lib/progressbar.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	test(`should correct maximum and minimum value`, () => {
+		progressbar.patch({min: 1, max: 5, value: 4});
+		expect(state).toContain({value: 4});
+
+		progressbar.patch({max: 3});
+		expect(state).toContain({min: 1, max: 3, value: 3});
+
+		progressbar.patch({max: 5});
+		expect(state).toContain({min: 1, max: 5, value: 3});
```
fbasso, line 51: max is a value that can be changed depending on the min (changing the min greater than the max or the max below the min). Maybe we should specify this behavior in a test too.

core/lib/progressbar.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+			slotContent: undefined,
+			slotDefault: undefined,
+			started: false,
+			striped: false,
+			value: 0,
+		});
+	});
+
+	test(`should calculate the percentage properly`, () => {
+		progressbar.patch({min: 1, max: 5, value: 4});
```
fbasso, line 37: In your test, you consider only a part of the state, meaning that a lot of use case may not be tested. For example, in this example (and in the UT of this file), the property `started` is not tested, but yet it changes from `true`to `false`.

Why not testing the full state directly ? 

For example

```
```suggestion
		let expectedState = state;
		progressbar.patch({min: 1, max: 5, value: 4});
		expectedState = {...expectedState, min: 1, max: 5, percentage: 75, value: 4, started: true};
		expect(state).toStrictEqual(expectedState);
		progressbar.patch({min: 10, max: 0});
		expectedState = {...expectedState, percentage: 0, min: 10, max: 10, value: 10, started: false, finished: true};
		expect(state).toStrictEqual(expectedState);
```

So that you'll be sure of all side effect of your api.

core/lib/progressbar.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+		expect(state).toContain({percentage: 75, value: 4});
+		progressbar.patch({min: 10, max: 0});
+		expect(state).toContain({percentage: 0, min: 10, max: 10, value: 10});
+	});
+
+	test(`should correct maximum and minimum value`, () => {
```
fbasso, line 43: Same remark as the previous test, testing the full state is more secure.

e2e/progressbar/progressbar.e2e-spec.ts
```diff
@@ -0,0 +1,36 @@
+import {ProgressbarPO} from '@agnos-ui/page-objects';
+import {expect, getTest} from '../fixture';
+import {ProgressbarDemoPO} from 'e2e/demo-po/progressbar.po';
+
+const test = getTest();
+test.describe(`Progressbar tests`, () => {
```
fbasso, line 6: Instead of testing the attributes individually, could you use a `getSate`  ? (see the `pagination.e2e-spec` example). The same idea than in the UT, we'll test more cases as all side effects of an action will be consider.

fbasso (changes requested):
Thanks for your previous changes @quentinderoubaix ! 

Some new suggestions here...

page-objects/lib/progressbar.po.ts
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	get locatorOuterBar(): Locator {
+		return this.locatorRoot.locator(this.selectors.outerBar);
+	}
+	get locatorInnerBar(): Locator {
+		return this.locatorRoot.locator(this.selectors.innerBar);
+	}
+
+	async getState(): Promise<State> {
```
fbasso, line 34: I don't know if we must call `getState`or just `state` (as the rating). It's a decision to take, but we need to be consistent...

react/demo/app/samples/progressbar/Striped.route.tsx
```diff
@@ -0,0 +1,21 @@
+import {Progressbar} from '@agnos-ui/react';
+
+const RatingDemo = () => (
```
fbasso, line 3: RatingDemo => ProgressbarDemo

page-objects/lib/progressbar.po.ts
```diff
@@ -0,0 +1,48 @@
...
+			const innerBar = rootNode.querySelector('.progress-bar');
+			return {
+				ariaLabel: rootNode.getAttribute('aria-label'),
+				ariaValueNow: rootNode.getAttribute('aria-valuenow'),
+				ariaValueMin: rootNode.getAttribute('aria-valuemin'),
+				ariaValueMax: rootNode.getAttribute('aria-valuemax'),
+				ariaValueText: rootNode.getAttribute('aria-valuetext'),
+				label: innerBar?.textContent?.trim(),
+				innerClasses: innerBar?.className?.trim()?.split(' ')?.sort() ?? [],
+			};
```
fbasso, line 45: I wonder if the height needs to be part of the state (?). Or maybe having the style attribute, as it depends on this property.

fbasso (changes requested):
Some little changes to do, it should be fine for the rest ;)

page-objects/lib/progressbar.po.ts
quentinderoubaix: sure, I will rename to state then

quentinderoubaix: note that I will not handle focustrack and transition pos that also currently use ```getState``` in this PR

page-objects/lib/progressbar.po.ts
quentinderoubaix: As discussed, I have added the outerHeight and innerWidth in the state, thanks !

fbasso (approved):

PR #115, opened 2023-08-28
Adding className as a common prop for all widgets

divdavem:


core/lib/commonProps.ts
```diff
@@ -0,0 +1,6 @@
+export interface AllWidgetsCommonPropsAndState {
```
fbasso, line 1: I think that just `WidgetsCommonPropsAndState` as a name is enough ;)

divdavem: Thank you for your suggestion. I have renamed the interface.

fbasso (approved):
Fine for me, thanks @divdavem !

divdavem:

PR #118, opened 2023-08-30
demo: add TOC on pages

fbasso:


demo/src/lib/layout/Section.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+	import Svg from './Svg.svelte';
+	import link from 'bootstrap-icons/icons/link-45deg.svg?raw';
+
+	export let level: 1 | 2 | 3;
```
divdavem, line 5: We could (probably in another PR) compute the level automatically based on the context.

demo/src/lib/stores.ts
```diff
@@ -1,6 +1,7 @@
 import {computed, get} from '@amadeus-it-group/tansu';
 import {browser} from '$app/environment';
 import {page} from '$app/stores';
+import {createIntersection} from '../../../core/lib/services/intersection';
```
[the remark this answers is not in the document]
divdavem: Also, we should add tests for the intersection utility, especially as it is exposed publicly (adding tests can be done in another PR).

demo/src/lib/layout/Section.svelte
fbasso: Yes, it can be a good idea. To be investigated and discussed in another PR indeed.

core/lib/services/intersection.ts
```diff
@@ -0,0 +1,77 @@
+import type {ReadableSignal} from '@amadeus-it-group/tansu';
+import {asReadable, derived} from '@amadeus-it-group/tansu';
+import {writablesForProps, type ValuesOrStores} from './stores';
+import {noop} from '../utils';
+
+export type IntersectionPropsConfig<T extends object> = ReadableSignal<Partial<T>> | ValuesOrStores<T>;
```
divdavem, line 6: Please do not duplicate this definition that already exists as `PropsConfig` in `stores.ts`

divdavem (approved):

fbasso:
Thanks @divdavem !

PR #121, opened 2023-09-04
Demo: ease main content layout

fbasso:
Remove the css grid layout in favor of responsive Bootstrap classes to better manage the different viewport sizes


quentinderoubaix (approved):
PR looks good to me.

While reviewing, I noticed that the toc and the sidebar are a bit different (font-weight of the title, top padding, padding between title and elements).
Do you think we should align those two displays ?

fbasso:
> While reviewing, I noticed that the toc and the sidebar are a bit different (font-weight of the title, top padding, padding between title and elements).
> Do you think we should align those two displays ?

Yes, I've done them independently. I'll have a look to see what it looks like if I change them.

By the way (and just for info), I didn't write it, but the purpose of this PR is not only to use the Bootstrap classes, but now we can have access to the mixin and variables. Which is useful in this case to get the Bootstrap breakpoints.

PR #123, opened 2023-09-08
fix: stackblitz with css

quentinderoubaix:


react/demo/app/samples/rating/Readonly.route.tsx
```diff
@@ -1,3 +1,4 @@
+import '@agnos-ui/common/samples/rating/custom.scss';
```
divdavem, line 1:
```suggestion
import '@agnos-ui/common/samples/rating/readonly.scss';
```

svelte/demo/samples/rating/Readonly.route.svelte
```diff
@@ -1,9 +1,10 @@
 <script lang="ts">
 	import {Rating} from '@agnos-ui/svelte';
 	import heartFill from 'bootstrap-icons/icons/heart-fill.svg?raw';
+	import '@agnos-ui/common/samples/rating/custom.scss';
```
divdavem, line 4:
```suggestion
	import '@agnos-ui/common/samples/rating/readonly.scss';
```

demo/src/lib/stackblitz/index.ts
```diff
@@ -5,21 +5,23 @@ import type {AsyncFilesSet, SampleInfo} from '../layout/sample';
 
 type StackblitzProcessor = (project: Project, sample: SampleInfo, framework: Frameworks) => Promise<void>;
 
+const SAMPLES_COMMON_ASSETS = /@agnos-ui\/common\/samples\/[^/]+/g;
+
+const cleanUpFile = (content: string) => {
+	return content.replace(SAMPLES_COMMON_ASSETS, '.');
```
divdavem, line 11: It would be better to do this during the build.
Note that this only applies when opening the sample in stackblitz, whereas it would be good to also see the simple local relative path ('./...' without '@agnos-ui/common/samples') when showing the source code in the demo site.

quentinderoubaix:
Thanks for the review @divdavem ! Have updated the PR following your recommendations.

demo/scripts/includeSamples.plugin.ts
```diff
@@ -73,7 +75,9 @@ export const includeSamples = (): Plugin => {
 					const frameworkFiles = files[framework];
 					output += `${JSON.stringify(framework)}:{entryPoint:${JSON.stringify(frameworkFiles[0].fileName)},files:{`;
 					frameworkFiles.forEach(({fileName, filePath}) => {
-						output += `[${JSON.stringify(fileName)}]: () => import(${JSON.stringify(filePath + '?raw')}).then(file=>file.default),`;
+						output += `[${JSON.stringify(fileName)}]: () => import(${JSON.stringify(
+							filePath + '?raw'
+						)}).then(file=>file.default).then(file=>file.replace(/@agnos-ui\\/common\\/samples\\/[^/]+/g,'.')),`;
```
divdavem, line 80: Unlike the previous code in your PR, this is now also applied when viewing the code in the demo site :+1: , but it is still done at runtime rather than at build time. I think it would be better to do it at built time.
To implement it, you can simply replace `+'?raw'` by something else of your choice (such as `+'?sampleCode'`) and process it in this plugin (in a way similar to [this](https://github.com/vitejs/vite/blob/c7969597caba80cf5d3348cba9f18ad9d14e9295/packages/vite/src/node/plugins/asset.ts#L183)). What do you think?

divdavem (approved):

PR #111, opened 2023-08-25
feat(accordion): demo examples

ValentinNelu:
Two examples for the accordion demo taken from ng-bootstrap

PR #136, opened 2023-09-13
demo: accordion sample simplification

fbasso:
The purpose is to simplify the accordion examples, to render the displayed code more readable (less text in the bodies).


ValentinNelu:
Actually with this [PR](https://github.com/AmadeusITGroup/AgnosUI/pull/111) we are already cleaning a lot the code, what do you think? @quentinderoubaix

fbasso:
No longer applicable

PR #140, opened 2023-09-14
feat(Demo): function props values select

ValentinNelu:
first draft, still need to check events props

### DIFF
--- a/common/propsValues.ts
+++ b/common/propsValues.ts
@@ -1,0 +1,46 @@
+import {collapseVerticalTransition, fadeTransition} from '../core/lib/transitions/bootstrap';
+import {getPaginationDefaultConfig, getRatingDefaultConfig} from '@agnos-ui/core';
+
+export type PropsValues = keyof typeof propsValuesList;
+
+const propsValuesList = {
+	//common
+	log: {value: (x: any) => console.log(x), label: 'Log'},
+	noop: {value: () => {}, label: 'Noop'},
+	//transitions
+	collapse: {value: collapseVerticalTransition, label: 'Collapse'},
+	fade: {value: fadeTransition, label: 'Fade'},
+	//rating
+	ratingValueFunction: {value: getRatingDefaultConfig().ariaValueTextFn, label: 'RatingValue'},
+	heartsRatingValue: {
+		value: (rating: number, maxRating: number) =>
+			`${rating} ${rating === 1 ? 'heart' : 'hearts'} out of ${maxRating} ${maxRating === 1 ? 'heart' : 'hearts'}`,
+		label: 'HeartsRatingValue',
+	},
+	starRatingValue: {
+		value: (rating: number, maxRating: number) =>
+			`${rating} ${rating === 1 ? 'star' : 'stars'} out of ${maxRating} ${maxRating === 1 ? 'star' : 'stars'}`,
+		label: 'StarRatingValue',
+	},
+	starIconRating: {value: getRatingDefaultConfig().slotStar, label: 'StarIconRating'},
+	heartIconRating: {value: ({fill}: {fill: number}) => String.fromCharCode(fill === 100 ? 9829 : 9825), label: 'HeartIconRating'},
+	//pagination
+	defaultPageFactory: {value: getPaginationDefaultConfig().pagesFactory, label: 'DefaultPageFactory'},
+	simpleLabel: {value: getPaginationDefaultConfig().ariaPageLabel, label: 'SimpleLabel'},
+	pageLabel: {value: (processPage: number, pageCount: number) => `Page ${processPage} of ${pageCount}`, label: 'PageLabel'},
+	slotNumberLabel: {value: getPaginationDefaultConfig().slotNumberLabel, label: 'SlotNumberLabel'},
+};
+
+export const getPropValues = (config: Record<string, any>) => {
+	const configUpdated = {...config};
+	for (const [key, value] of Object.entries(configUpdated)) {
+		if (propsValuesList[value?.value as PropsValues] !== undefined) {
+			configUpdated[key] = propsValuesList[value.value as PropsValues].value;
+		}
+	}
+	return configUpdated;
+};
+
+export const getPropValueLabel = (key: PropsValues) => {
+	return propsValuesList[key as PropsValues].label;
+};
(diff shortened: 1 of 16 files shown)


angular/demo/src/app/samples/accordion/playground.route.ts
```diff
@@ -27,9 +28,9 @@ export default class PlaygroundComponent {
 
 	constructor() {
 		hashChangeHook((props) => {
-			this.widget?._widget.patch({...undefinedAccordionConfig, ...props});
+			this.widget?._widget.patch(getPropValues({...undefinedAccordionConfig, ...props}));
```
divdavem, line 31: Is calling `getPropValues` really useful here and line 33? Isn't it already called in `utils.ts`?

common/propsValues.ts
```diff
@@ -0,0 +1,38 @@
...
+	starIconRating: getRatingDefaultConfig().slotStar,
+	heartIconRating: ({fill}: {fill: number}) => String.fromCharCode(fill === 100 ? 9829 : 9825),
+	//pagination
+	defaultPageFactory: getPaginationDefaultConfig().pagesFactory,
+	simpleLabel: getPaginationDefaultConfig().ariaPageLabel,
+	pageLabel: (processPage: number, pageCount: number) => `Page ${processPage} of ${pageCount}`,
+	slotNumberLabel: getPaginationDefaultConfig().slotNumberLabel,
+};
+
+export const getPropValues = (config: Record<string, any>) => {
```
divdavem, line 31: I think `getPropValues` should not modify its argument. It is better to do a copy before doing the change.
Otherwise, it should be clear in its name that it mutates the argument, but then we should not call it with something we directly got from dependency injection.

common/propsValues.ts
```diff
@@ -0,0 +1,38 @@
...
+const propsValuesList: Record<string, any> = {
+	//common
+	log: (x: any) => console.log(x),
+	noop: () => {},
+	//transitions
+	collapse: collapseVerticalTransition,
+	fade: fadeTransition,
+	//rating
+	ratingValue: getRatingDefaultConfig().ariaValueTextFn,
+	hartsRatingValue: (rating: number, maxRating: number) =>
```
divdavem, line 18:
```suggestion
	heartsRatingValue: (rating: number, maxRating: number) =>
```

common/propsValues.ts
```diff
@@ -0,0 +1,38 @@
...
+		`${rating} ${rating === 1 ? 'heart' : 'hearts'} out of ${maxRating} ${maxRating === 1 ? 'heart' : 'hearts'}`,
+	startRatingValue: (rating: number, maxRating: number) =>
```
divdavem, line 20:
```suggestion
	starRatingValue: (rating: number, maxRating: number) =>
```

demo/src/routes/[framework]/components/accordion/playground/+page.svelte
```diff
@@ -3,6 +3,33 @@
 	import {getAccordionDefaultConfig} from '@agnos-ui/core';
 	import sample from '@agnos-ui/samples/accordion/playground';
 	import doc from '../../../../../../../core/dist/api.json?accordion&extractApi';
+	import type {SelectValues} from '@agnos-ui/common/propsValues';
+	export let listPropsValues: {[key in keyof ReturnType<typeof getAccordionDefaultConfig>]?: SelectValues[]} = {
+		onShown: [
+			{value: 'noop', label: 'Noop'},
```
Comment (divdavem) [inline demo/src/routes/[framework]/components/accordion/playground/+page.svelte:9]: Instead of repeating everywhere the value and label, wouldn't it be better to accept here only the value, and keep the label in a central place (along with the function definition)?

demo/src/lib/layout/playground/PlaygroundValue.svelte
```diff
@@ -2,13 +2,17 @@
...
 
 	export let type: string;
 	export let defaultValue: any;
 	export let api: SingleValueContextApi;
 	export let placeholder = '';
 	export let ariaLabel: string;
+	export let selectValues: SelectValues[] = [];
 
+	// used only for the select where the api. value = {value: [key]}
+	let value: string | undefined = api.isEmpty ? undefined : api.value.value;
```
divdavem, line 15: I think this logic should be managed in the api.

demo/src/lib/layout/playground/PlaygroundValue.svelte
```diff
@@ -26,38 +30,47 @@
 			/>
 		</div>
 	</td>
-	<td class="checkbox"
-		><button class="btn btn-link icon" title="Clear value" disabled={api.isEmpty} on:click={api.clear}
+	<td class="checkbox align-middle"
+		><button class="btn btn-link m-0 p-0 d-flex mx-auto" title="Clear value" disabled={api.isEmpty} on:click={api.clear}
 			><Svg className="icon-20" svg={clearIcon} /></button
 		></td
 	>
```
divdavem, line 37: Shouldn't this `<td>` with the `clearIcon` factorized instead of being repeated for each type of field?

ValentinNelu: It is from old PR I could take care of it in the same PR I guess

demo/src/routes/[framework]/components/accordion/playground/+page.svelte
Comment (ValentinNelu) [reply demo/src/routes/[framework]/components/accordion/playground/+page.svelte:9]: Indeed makes sense since there are lot of reused labels

demo/src/lib/layout/playground/PlaygroundValue.svelte
divdavem: Indeed, it was already repeated before your PR, but as you did the same change several times (for each copy), maybe you can improve the code at the same time.

angular/demo/src/app/samples/accordion/playground.route.ts
ValentinNelu: you are right, I forgot to remove it

common/propsValues.ts
```diff
@@ -0,0 +1,46 @@
...
+	const configUpdated = {...config};
+	for (const [key, value] of Object.entries(configUpdated)) {
+		if (propsValuesList[value?.value as PropsValues] !== undefined) {
+			configUpdated[key] = propsValuesList[value.value as PropsValues].value;
+		}
+	}
+	return configUpdated;
+};
+
+export const getPropValueLabel = (key: string) => {
```
divdavem, line 44:
```suggestion
export const getPropValueLabel = (key: PropsValues) => {
```

divdavem (approved):
After changing the last small things I have mentioned, I think you can merge the PR. Thank you! :+1:

demo/src/lib/layout/playground/PlaygroundValue.svelte
divdavem: I saw you changed this, thank you!
But I was thinking in a more complex change that should probably be done in a separate PR (to better use reactivity).

PR #175, opened 2023-10-04
feat: Add typeArray and allow passing equal function to bindableDerived

MarkoOleksiyenko:


core/lib/services/checks.ts
```diff
@@ -34,6 +34,15 @@ export function isString(value: any): value is string {
 	return typeof value === 'string';
 }
 
+/**
+ * an array type guard
+ * @param value the value to check
+ * @returns true if the value is an array
+ */
+export function isArray(value: any) {
```
divdavem, line 42:
```suggestion
export function isArray(value: any): value is any[] {
```
(cf https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

divdavem, line 42: Or maybe we should do:
```ts
export const isArray = Array.isArray;
```

core/lib/services/stores.spec.ts
```diff
@@ -442,5 +444,35 @@ describe(`Stores service`, () => {
...
+			value$.subscribe((value) => values.push(value));
+			expect(values).toEqual([[1]]);
+
+			dirtyValue$.set([1]); // no change
+			expect(onChangeCalls).toEqual([]);
+			expect(values).toEqual([[1]]);
+
+			dirtyValue$.set([2.5]);
+			expect(onChangeCalls).toEqual([[2]]);
+			expect(values).toEqual([[1], [2]]);
```
divdavem, line 471: Maybe you could also check that the value of `dirtyValue$` changed as well (and is now `[2]`), and also at the end of the test (for `[5,7]`).

core/lib/services/writables.ts
```diff
@@ -23,3 +23,14 @@ export const typeFunction: WritableWithDefaultOptions<(...args: any[]) => any> =
 	normalizeValue: testToNormalizeValue(isFunction),
 	equal: Object.is,
 };
+
+export const typeArray: WritableWithDefaultOptions<any[]> = {
+	normalizeValue: testToNormalizeValue(isArray),
+	equal: (a, b) => {
```
divdavem, line 29: You should add a test for this equal function.

core/lib/services/writables.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+		// set the wrong value
+		store$.set(100 as any);
+		expect(consoleErrorSpy).toHaveBeenCalledWith('Not setting invalid value', 100);
+		// the store is not updated with the wrong value
+		expect(store$()).toStrictEqual([0]);
+
+		const testArray = [15, 20];
+		store$.set(testArray);
+
+		expect(store$()).toStrictEqual([15, 20]);
```
divdavem, line 26:
```suggestion
		expect(store$()).toBe(testArray);
```

core/lib/services/writables.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+		testArray[1] = 15;
+		store$.set(testArray);
+
+		expect(store$()).toStrictEqual([15, 15]);
+	});
+
+	test(`typeArray equal function should compare values`, () => {
```
divdavem, line 34: To fix typings, you can use `typeArray.equal!` instead of  `typeArray.equal` in all those lines (or do `const equal = typeArray.equal!` and use `equal` in the following lines).

core/lib/services/writables.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+		expect(typeArray.equal([15, 15], [15, 15])).toBe(true);
+		expect(typeArray.equal([15, 15], [15, 10])).toBe(false);
+		const arr = [15, 15];
+		const arr2 = arr;
+		arr2[1] = 10;
+		expect(typeArray.equal(arr, arr2)).toBe(true);
+		expect(typeArray.equal([15], [15, 15])).toBe(false);
+		expect(typeArray.equal([15, 15], [15])).toBe(false);
+		expect(typeArray.equal(undefined, [15])).toBe(false);
```
divdavem, line 43: `undefined` is not part of the accepted types for this equal function.
You should do `undefined as any` if you still want to test this case.

core/lib/services/writables.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+		expect(consoleErrorSpy).toHaveBeenCalledWith('Not setting invalid value', 100);
+		// the store is not updated with the wrong value
+		expect(store$()).toStrictEqual([0]);
+
+		const testArray = [15, 20];
+		store$.set(testArray);
+
+		expect(store$()).toStrictEqual([15, 20]);
+
+		testArray[1] = 15;
```
divdavem, line 28: I am not really in favor of using stores this way. I prefer immutability, otherwise we need `equal` functions that consider objects as always different. I would remove this part of the test.

core/lib/services/writables.spec.ts
MarkoOleksiyenko: true, will type the method params

core/lib/services/writables.ts
```diff
@@ -23,3 +23,17 @@ export const typeFunction: WritableWithDefaultOptions<(...args: any[]) => any> =
 	normalizeValue: testToNormalizeValue(isFunction),
 	equal: Object.is,
 };
+
+export const typeArray: WritableWithDefaultOptions<any[]> = {
+	normalizeValue: testToNormalizeValue(isArray),
+	equal: (a: any[], b: any[]) => {
```
divdavem, line 29:
```suggestion
	equal: (a, b) => {
```
(the type of equal is already implied by `WritableWithDefaultOptions<any[]>` line 27)

MarkoOleksiyenko: it is, however, the IDE highlights it in red saying that it as implicit type

divdavem (approved):

PR #182, opened 2023-10-10
feat(slider): Add slider component

MarkoOleksiyenko:


core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+
+	/**
+	 * Dom Rectangle of the minimum label
+	 */
+	minLabelDomRect: DOMRect;
+
+	/**
+	 * Dom Rectangle of the maximum label
+	 */
+	maxLabelDomRect: DOMRect;
```
divdavem, line 55: Do you use `sliderDomRect`, `minLabelDomRect` or `maxLabelDomRect` outside of the core?
As far as I can see, it is not useful to expose them in the state, they could be part of `SliderProps` instead of `SliderCommonPropsAndState`.

divdavem: Actually, instead of using props, maybe it would be better to have directives for each of those DOM elements, and then observing the size of the elements could be done in the core, and that would avoid repeating this logic in other frameworks.

maxokorokov: It would be nice to avoid duplicating resize observer logic across frameworks indeed

maxokorokov: But then again we could see it when implementing React or Svelte version?

core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+	positionTop: typeArray,
+	onMouseDown: typeFunction,
+	onKeyDown: typeFunction,
+};
+
+const sliderHandleProps = Object.keys(defaultHandleConfig) as (keyof SliderHandleProps)[];
+
+function createSliderHandle(config?: PropsConfig<SliderHandleProps>): SliderHandleWidget {
+	const [{onMouseDown$, onKeyDown$, ...stateProps}, patch] = writablesForProps(defaultHandleConfig, config, configSliderHandleValidator);
+	const prevCoordinate$ = writable(-1);
```
divdavem, line 549: Is this used anywhere?

core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+
+	/**
+	 * Sorted slider values
+	 */
+	sortedValues: number[];
+
+	/**
+	 * Mix label left offset in %
+	 */
+	mixLabelLeft: number;
```
maxokorokov, line 92: I don't understand neither naming nor documentation for these. What does it do?

core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+
+	/**
+	 * Array of the top offsets for the tooltips in %
+	 */
+	handleTooltipTop: number[];
+
+	/**
+	 * If `visible` the minimum label will be displayed
+	 */
+	minValueLabelDisplay: string;
```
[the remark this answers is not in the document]
maxokorokov: Also maybe `minValueLabelVisible` or `minValueLabelDisplayed` or similar in this case?

core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+
+	/**
+	 * If `visible` the maximum label will be displayed
+	 */
+	maxValueLabelDisplay: string;
+
+	/**
+	 * If `visible` the label when the handles are close is displayed
+	 */
+	mixLabelDisplay: string;
```
maxokorokov, line 122: Maybe `combinedLabel...`, `stichedLabel...`, `squashedLabel...` ?

core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+
+	/**
+	 * Array of the sorted handles to display
+	 */
+	sortedHandleDisplay: {value: number; id: number}[];
```
maxokorokov, line 127: maybe just `sortedHandles` ?

core/lib/slider.ts
```diff
@@ -0,0 +1,569 @@
...
+
+	/**
+	 * Array containing all the slider handles
+	 */
+	handlesWidget: SliderHandleWidget[];
```
maxokorokov, line 132: `handleWidgets` ?

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,539 @@
...
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	host: {
+		class: 'au-slider-handle',
+		'[class]': "state().vertical ? 'au-slider-handle-vertical' : 'au-slider-handle-horizontal'",
+		'[style.left.%]': 'state().positionLeft[state().handleId]',
+		'[style.top.%]': 'state().positionTop[state().handleId]',
+		'(keydown)': 'widget.actions.onKeyDown($event, state().handleId)',
+		'(mousedown)': 'widget.actions.onMouseDown($event)',
+	},
+	hostDirectives: [
```
maxokorokov, line 42: `hostDirectives: [UseDirective]`

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,539 @@
...
+			[attr.disabled]="state().disabled ? true : null"
+			[style.left.%]="state().progressLeft[i]"
+			[style.bottom.%]="state().progressBottom[i]"
+			[style.width.%]="state().progressWidth[i]"
+			[style.height.%]="state().progressHeight[i]"
+		></div>
+		<div
+			#minLabel
+			[class]="state().vertical ? 'au-slider-label-vertical au-slider-label-vertical-min' : 'au-slider-label au-slider-label-min'"
+			[style.visibility]="state().minValueLabelDisplay"
```
maxokorokov, line 183: By the way why is it hidden and not removed from DOM?

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,539 @@
...
+	 * SliderWidget hold the state and actions applicable to the auSliderComponent
+	 */
+	readonly widget = toSlotContextWidget(this._widget);
+	readonly state: Signal<SliderState> = toSignal(this._widget.state$, {requireSync: true});
+	readonly api = this._widget.api;
+
+	/**
+	 * Dom Rectangle of the slider
+	 */
+	@Input('auSliderDomRect') sliderDomRect: DOMRect | undefined;
```
maxokorokov, line 390: Why is is a component input?

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,539 @@
...
+	private _prevCoordinate = -1;
+
+	// TODO how to declare it in props?
+	// eslint-disable-next-line @agnos-ui/angular-check-props
+	@Input('auSliderActions') sliderActions: SliderActions;
+
+	/**
+	 * The id of the slider handle.
+	 */
+	@Input('auHandleId') handleId: number | undefined;
```
maxokorokov, line 62: Why have all these inputs? Everyhing comes from the state, no?

core/lib/slider.ts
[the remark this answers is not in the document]
maxokorokov: Also why changing visibility and not removing it from DOM ? (I think I left similar comment below.

core/lib/slider.ts
maxokorokov: Also the whole prevCoordinate handling and it being a part of the state looks... hmm... fishy :)

core/lib/slider.ts
MarkoOleksiyenko: agree, it is not used outside the core, only updated from there, it should be only a prop

core/lib/slider.ts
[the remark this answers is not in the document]
MarkoOleksiyenko: I am not removing it from the dom as the resizeobserver looks for the element in the DOM, so if we remove it will be undefined and will break the computations
The strings are used because the style accepts 'hidden' or 'visible' and not the true/false, so to avoid conversion I chose strings

core/lib/slider.ts
divdavem: Actually, as I said in my second comment, it should not even be a prop, because it does not really make sense to correspond to an `@Input` in the angular component. So having an internal writable in the core that is updated by an exposed directive that is put on the corresponding DOM elements seems more appropriate to me.

core/lib/slider.ts
MarkoOleksiyenko: oh, yes, forgot to remove it from the state (it is only used in the angular handle implementation, so should not be here)

core/lib/slider.ts
MarkoOleksiyenko: combinedLabel sounds better indeed :)

core/lib/slider.ts
MarkoOleksiyenko: will add the 'position' to the variable name

core/lib/slider.ts
MarkoOleksiyenko: yeah, 'display' can be omitted. We need to have this array here for the ngForOf in the template, as the registerHandle only works from inside the component when it is added in DOM, so initially the array of widgets is empty and cannot be used to display and trackby the handles

angular/lib/src/lib/slider/slider.component.ts
MarkoOleksiyenko: oh, i thought I removed this directive, no need for it actually

angular/lib/src/lib/slider/slider.component.ts
MarkoOleksiyenko: the reziseObserver always needs it in the dom to return proper value for the DOMRect otherwise the computations fail

core/lib/slider.ts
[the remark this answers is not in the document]
divdavem: You already have a conversion from boolean to string in the core, just move it to the template if you really need to use the style.
But if you move the ResizeObserver to the core as I am advising you, you will have to handle the case of missing DOM elements anyway and you could as well remove the elements from the DOM instead of hiding them.

angular/lib/src/lib/slider/slider.component.ts
MarkoOleksiyenko: from my understanding there should be input for every prop, otherwise it doesn't build properly

angular/lib/src/lib/slider/slider.component.ts
MarkoOleksiyenko: if these inputs are not here, the build doesn't pass and I don't think that putting the ignore is the best idea

angular/lib/src/lib/slider/slider.component.ts
divdavem: It should not be a prop

angular/lib/src/lib/slider/slider.component.ts
divdavem: I don't think the slider handle should be a widget.

MarkoOleksiyenko: it is a widget because we need to pass the implementation on the onMouseDown and onKeydown per framework, so it is a prop, otherwise I don't see how we can override it

core/lib/slider.ts
maxokorokov: nah, either you should add `offset` or `position` to all props or to none. My comment was mostyl about `mixLabel` part.

angular/lib/src/lib/slider/slider.component.ts
maxokorokov: even better :)

angular/lib/src/lib/slider/slider.component.ts
maxokorokov: Well then it makes sense to me have the `visible` / `hidden` and not boolean

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,393 @@
...
+	 * Method that holds all the necessary emits to update the input value of the slider
+	 */
+	emitControlValues() {
+		this.onChange(this.state().sortedValues);
+		this.onTouched();
+	}
+
+	sliderClick($event: any) {
+		this._widget.actions.updateSliderSize();
+		this.widget.actions.click(this.state().vertical ? $event.clientY : $event.clientX);
```
divdavem, line 380: I think you are doing here what should be done in the core.
Remove the `sliderClick` method here, and just call directly `widget.actions.click($event)` in the template.

core/lib/slider.ts
```diff
@@ -0,0 +1,509 @@
...
+	 */
+	updateSliderSize(): void;
+
+	/**
+	 * Method describing the behavior of the slider handle on mouse down event
+	 * @param event - mouse event
+	 * @param handleId - numeric id of the handle
+	 */
+	onMouseDown(event: MouseEvent, handleId: number): void;
+}
```
divdavem, line 167: You should remove the "on" prefix (`mouseDown`, `keydown`) to make actions consistent (with `click` and with other widgets).

core/lib/slider.ts
```diff
@@ -0,0 +1,509 @@
...
+	 * Method to process the keyboard event
+	 * @param event - keyboard event object
+	 * @param handleNumber - id of the modified handle
+	 */
+	onKeydown(event: KeyboardEvent, handleNumber: number): void;
+
+	/**
+	 * Method to manually invoke the slider size recomputation
+	 */
+	updateSliderSize(): void;
```
divdavem, line 159: Do you actually need this anywhere (after moving the click event handler to the core)?

angular/lib/src/lib/slider/slider.component.ts
MarkoOleksiyenko: agree at this point we don't need the additional handling outside the core

angular/demo/src/app/samples/slider/range.route.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Component({
+	standalone: true,
+	imports: [AgnosUIAngularModule, ReactiveFormsModule, FormsModule],
+	providers: [
+		provideWidgetsConfig((config) => {
+			config.slider = {...config.slider};
+			return config;
+		}),
+	],
```
divdavem, line 13: You can remove line 8-13 also here.

angular/demo/src/app/samples/slider/vertical.route.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Component({
+	standalone: true,
+	imports: [AgnosUIAngularModule, ReactiveFormsModule, FormsModule],
+	providers: [
+		provideWidgetsConfig((config) => {
+			config.slider = {...config.slider};
+			return config;
+		}),
+	],
```
divdavem, line 13: You can remove line 8-13 also here.

e2e/samplesMarkup.e2e-spec.ts
```diff
@@ -30,6 +30,10 @@ test.describe.parallel(`Samples markup consistency check`, () => {
 			if (route === 'accordion/playground' && testInfo.project.name.startsWith('angular:')) {
 				test.skip();
 			}
+			// skipping consistency checks for react and svelte for slider
+			if (testInfo.project.name.startsWith('react:') || testInfo.project.name.startsWith('svelte:')) {
```
divdavem, line 34: Where do you check that it is for the slider? Don't disable samples markup check in react and svelte for all widgets!

demo/src/routes/[framework]/components/slider/+layout.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+	import ComponentPage from '../../../../lib/layout/ComponentPage.svelte';
```
Comment (divdavem) [inline demo/src/routes/[framework]/components/slider/+layout.svelte:2]:
```suggestion
	import ComponentPage from '$lib/layout/ComponentPage.svelte';
```

demo/src/routes/[framework]/components/slider/api/+page.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+	import {getAlertDefaultConfig} from '@agnos-ui/core';
+	import widgetDoc from '../../../../../../../core/dist/api.json?slider&extractApi';
+	import ApiDoc from '../../../../../lib/layout/doc/ApiDoc.svelte';
```
Comment (divdavem) [inline demo/src/routes/[framework]/components/slider/api/+page.svelte:4]:
```suggestion
	import ApiDoc from '$lib/layout/doc/ApiDoc.svelte';
```

e2e/samplesMarkup.e2e-spec.ts
MarkoOleksiyenko: oh, missed the line with the route :D

core/lib/slider.ts
```diff
@@ -0,0 +1,500 @@
...
+	 * Method to handle click on the slider
+	 * @param event - mouse event
+	 */
+	click(event: MouseEvent): void;
+	/**
+	 * Method to process the keyboard event
+	 * @param event - keyboard event object
+	 * @param handleNumber - id of the modified handle
+	 */
+	onKeydown(event: KeyboardEvent, handleNumber: number): void;
```
divdavem, line 153:
```suggestion
	keydown(event: KeyboardEvent, handleNumber: number): void;
```

angular/demo/src/app/samples/slider/default.route.ts
```diff
@@ -0,0 +1,57 @@
...
+import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
+
+@Component({
+	standalone: true,
+	imports: [AgnosUIAngularModule, ReactiveFormsModule, FormsModule],
+	providers: [
+		provideWidgetsConfig((config) => {
+			config.slider = {...config.slider};
+			return config;
+		}),
```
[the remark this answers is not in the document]
divdavem: You removed those lines in the 2 other samples (thank you!) but you seem to have forgotten this one!

core/lib/slider.ts
MarkoOleksiyenko: thought that I've committed the change, now should be good!

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+	host: {
+		class: 'au-slider',
+		'[class]': 'state().vertical ? "au-slider-vertical" : "au-slider-horizontal"',
+		'(blur)': 'handleBlur()',
+	},
+	template: `
+		<div [class]="state().vertical ? 'au-clickable-slider-area-vertical' : 'au-clickable-slider-area'" (click)="sliderClick($event)"></div>
+		<div
+			*ngFor="let repeat of [].constructor(state().values.length - 1 || 1); let i = index"
+			#progress
```
divdavem, line 39: This name is not used anywhere.

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+			#minLabel
+			[class]="state().vertical ? 'au-slider-label-vertical au-slider-label-vertical-min' : 'au-slider-label au-slider-label-min'"
+			[style.visibility]="state().minValueLabelDisplay"
+			[attr.disabled]="state().disabled ? true : null"
+			[auUse]="widget.directives.minLabelDirective"
+		>
+			{{ state().min }}
+		</div>
+		<div
+			#maxLabel
```
divdavem, line 57: This name is not used anywhere.

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+			#progress
+			class="au-slider-progress"
+			[attr.disabled]="state().disabled ? true : null"
+			[style.left.%]="state().progressLeft[i]"
+			[style.bottom.%]="state().progressBottom[i]"
+			[style.width.%]="state().progressWidth[i]"
+			[style.height.%]="state().progressHeight[i]"
+		></div>
+		<div
+			#minLabel
```
divdavem, line 48: This name is not used anywhere.

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+	private _sliderDirective: ReturnType<Directive>;
+	private _zone = inject(NgZone);
+	// needed to avoid changedetection on the mouse move over the same coordinate (vertical or horizonal movement)
+	private _prevCoordinate = -1;
+
+	/**
+	 * Minimum label element reference
+	 */
+	@ViewChild('minLabel')
+	private _minLabelRef: ElementRef;
```
divdavem, line 240: This is not used anywhere.

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+	/**
+	 * Maximum label element reference
+	 */
+	@ViewChild('maxLabel')
+	private _maxLabelRef: ElementRef;
```
divdavem, line 245: This is not used anywhere.

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+})
+export class SliderComponent implements OnChanges, AfterViewInit, OnDestroy {
+	/**
+	 * auSlider element reference
+	 */
+	private _elementRef = inject(ElementRef);
+	private _sliderDirective: ReturnType<Directive>;
+	private _zone = inject(NgZone);
+	// needed to avoid changedetection on the mouse move over the same coordinate (vertical or horizonal movement)
+	private _prevCoordinate = -1;
```
divdavem, line 234: This is not used anywhere.

angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -0,0 +1,385 @@
...
+	@Input('auVertical')
+	vertical: boolean | undefined;
+
+	/**
+	 * An event emitted when slider values are changed
+	 *
+	 * Event payload equals to the updated slider values
+	 */
+	// TODO remove the disable when angular-check-props fixed
+	// eslint-disable-next-line @agnos-ui/angular-check-props
```
divdavem, line 318: I don't see any issue when removing this `eslint-disable-next-line` line.

MarkoOleksiyenko: hmm, did the npm install and build and it underlines it saying that it is not overridden in the constructor, mb the VS targets the wrong thing again

core/lib/slider.ts
```diff
@@ -0,0 +1,500 @@
...
+	);
+	const updateSliderSize$ = writable({});
+	const sliderDomRect$ = computed(
+		() => {
+			sliderResized$();
+			updateSliderSize$();
+			return sliderDom$()?.getBoundingClientRect() ?? ({} as DOMRect);
+		},
+		{
+			equal: (a, b) => Object.is(a, b),
```
divdavem, line 263:
```suggestion
			equal: Object.is,
```

core/lib/slider.spec.ts
```diff
@@ -0,0 +1,654 @@
...
+import {createSlider} from './slider';
+import type {WidgetState} from './types';
+
+// TODO move to the utils?
+function keyboardEvent(key: string): KeyboardEvent {
+	return {
+		key,
+		preventDefault() {},
+		stopPropagation() {},
+	} as KeyboardEvent;
```
divdavem, line 14: `return new KeyboardEvent('keydown', {key})`

core/lib/slider.spec.ts
```diff
@@ -0,0 +1,654 @@
...
+		const minLabel = document.createElement('div');
+		vi.spyOn(minLabel, 'getBoundingClientRect').mockReturnValue({x: 10, y: 5, width: 3, height: 4} as DOMRect);
+		const maxLabel = document.createElement('div');
+		vi.spyOn(maxLabel, 'getBoundingClientRect').mockReturnValue({x: 100, y: 5, width: 3, height: 4} as DOMRect);
+
+		defConfig = writable({});
+		slider = createSlider({config: computed(() => ({...defConfig()}))});
+		slider.directives.sliderDirective(sliderElement);
+		slider.directives.minLabelDirective(minLabel);
+		slider.directives.maxLabelDirective(maxLabel);
```
divdavem, line 60: It would be good to store the return value from the directives to call `.destroy()` on them at the end of each test.

core/lib/slider.spec.ts
```diff
@@ -0,0 +1,654 @@
...
+		unsubscribe = slider.state$.subscribe((newState) => {
```
divdavem, line 61: Note that `beforeEach` can return an `afterEach` function, which would allow you to store the `unsubscribe` variable as a `const` locally in the `beforeEach` function.

e2e/slider/slider.e2e-spec.ts
```diff
@@ -0,0 +1,374 @@
...
+		disabled: null,
+		style: 'left: 0%; top: 60%;',
+	},
+];
+
+const test = getTest();
+test.describe(`Slider tests`, () => {
+	test.describe(`Basic slider`, () => {
+		test(`should snap the handle to correct tick on the slider click event`, async ({page}, testInfo) => {
+			if (testInfo.project.name.startsWith('react:') || testInfo.project.name.startsWith('svelte:')) {
```
divdavem, line 61: Instead of repeating this condition in each test, you can use `test.skip` in `describe`, as documented here:
https://playwright.dev/docs/test-annotations#conditionally-skip-a-group-of-tests

divdavem: I have actually done it in this commit: https://github.com/AmadeusITGroup/AgnosUI/commit/cf480e7cee5b1cc812fa50a8febe14c39294bf93

divdavem (approved):

PR #195, opened 2023-10-25
test: run core and angular tests in browser instead of node.js

divdavem:


code-coverage/lib/vitestProvider.ts
```diff
@@ -15,10 +13,12 @@ const customCoverageProviderModule: CoverageProviderModule = {
 				return ctx!.config.coverage;
 			},
 			async clean(clean?: boolean) {
+				const {default: setup} = await import('./setup');
```
quentinderoubaix, line 16: The two imports (setup and reportCoverage) could be moved to line 7, before the return.

quentinderoubaix (approved):

fbasso (approved):
LGTM ;)

code-coverage/lib/vitestProvider.ts
divdavem: Indeed, that would be possible as well.

PR #217, opened 2023-11-06
feat: add modal sample using already defined component

quentinderoubaix:


angular/lib/src/lib/modal/modal.component.ts
```diff
@@ -213,25 +213,35 @@ export class ModalComponent implements OnChanges, AfterContentChecked {
...
 
-	@Input('auSlotStructure') slotStructure: SlotContent<ModalContext>;
+	// eslint-disable-next-line @agnos-ui/angular-check-props
+	@Input('auSlotStructure') slotStructure: SlotContent<ModalContext<Data>>;
 	@ContentChild(ModalStructureDirective, {static: false})
-	slotStructureFromContent: ModalStructureDirective | null;
+	slotStructureFromContent: ModalStructureDirective<Data> | null;
 
-	@Input('auSlotHeader') slotHeader: SlotContent<ModalContext>;
+	// eslint-disable-next-line @agnos-ui/angular-check-props
```
quentinderoubaix, line 221: @divdavem the angular-check-props plugin was failing and am not really sure why...

quentinderoubaix: Thx !

react/lib/modal/Modal.tsx
```diff
@@ -1,70 +1,76 @@
...
-			</Portal>
-		</>
+				</div>
+			)}
+		</Portal>
 	);
-});
+}
+
+export const Modal = forwardRef(ModalInner) as <Data>(
```
divdavem, line 74: Why did you introduce the `ModalInner` function instead of keeping the concise inline syntax?
Why did you change the arrow function syntax (in all functions in this file) to use the more verbose function syntax?

react/lib/modal/Modal.tsx
```diff
@@ -1,70 +1,76 @@
...
+	props: PropsWithChildren<Partial<ModalProps<Data>>> & {ref?: Ref<unknown> | undefined}
```
divdavem, line 75: `Ref<unknown>` is suspicious to me. Why don't you use the correct type `Ref<ModalWidget<Data>['api']>`?

divdavem (changes requested):
Thank you @quentinderoubaix for your PR.
It is already very good but it would be even better if you could fix the few things I have commented.

react/lib/modal/Modal.tsx
quentinderoubaix: For the ```ModalInner```, I followed a recommendation (https://fettblog.eu/typescript-react-generic-forward-refs/) as forwardRef and generic component don' t work well together.
I will put back the other functions to arrow syntax

react/lib/modal/Modal.tsx
quentinderoubaix: Thx for the help, it is not correctly typed.

react/lib/modal/Modal.tsx
quentinderoubaix: ModalInner removed too.

demo/scripts/includeSamples.plugin.ts
```diff
@@ -54,7 +54,11 @@ export const includeSamples = (): Plugin => {
 						for (const dependency of dependencies) {
 							const dependencyParts = dependency.split('/');
 							if (dependencyParts[0] === '.') {
-								await addFile(framework, path.basename(dependency), path.join(directory, dependency));
+								let relativePath = dependency;
+								if (dependency.endsWith('.component') && framework === 'angular') {
```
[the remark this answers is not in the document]
quentinderoubaix: Condition remade to be framework independant.

quentinderoubaix:
Have made the proposed changes, thx for the review @divdavem !

divdavem (approved):
Thank you! :+1:

PR #218, opened 2023-11-06
feat(style-bootstrap): slider style refactor + style-bootstrap package

ValentinNelu:


angular/lib/src/lib/slider/slider.component.ts
```diff
@@ -70,9 +67,9 @@ import {take} from 'rxjs';
 				[attr.aria-valuemax]="state().max"
 				[attr.aria-readonly]="state().readonly ? true : null"
 				[attr.aria-disabled]="state().disabled ? true : null"
+				[attr.tabindex]="state().disabled ? -1 : null"
 				[attr.aria-valuenow]="item.value"
 				[attr.aria-valuetext]="item.value"
-				[attr.disabled]="state().disabled ? true : null"
```
divdavem, line 75: You should not remove the disabled attribute on the `<button>` element (and you should not add tabindex="-1").
Note that, unlike the disabled attribute, tabindex="-1" does not prevent the ability to focus the button by click.

divdavem (approved):

PR #221, opened 2023-11-08
Slider: Svelte and React version

fbasso:


core/lib/services/writables.ts
```diff
@@ -7,10 +7,28 @@ export const testToNormalizeValue =
...
 };
 
+/**
+ * A factory function that creates a type guard function to check and rectify a value is within a specified range.
+ *
+ * @param min - The minimum value allowed.
+ * @param max - The maximum value allowed.
+ * @returns A type guard function that returns the clamp value if the value is a value number, and INVALID_VALUE otherwise.
+ */
+export function typeNumberInRangeFactory(min: number, max: number) {
```
quentinderoubaix, line 23: Quite useful new guard !
I guess we need to use it many of our current numbers (pagination collectionSize / pageSize / size, rating maxRating... etc)
Do you want to do this in this PR or create a different issue ?

angular/demo/src/app/samples/slider/vertical.route.ts
```diff
@@ -6,15 +6,14 @@ import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
 	standalone: true,
 	imports: [AgnosUIAngularModule, ReactiveFormsModule, FormsModule],
 	template: `
-		<h2>Vertical slider</h2>
 		<div class="d-flex" style="height: 350px">
 			<div class="col-6" style="height: 300px">
-				<au-component auSlider [auMin]="0" [auMax]="100" [auStepSize]="1" [auVertical]="true" [formControl]="sliderControl"></au-component>
-				Form control values: {{ sliderControl.value?.join(', ') }}
+				<div auSlider [auMin]="0" [auMax]="100" [auStepSize]="1" [auVertical]="true" [formControl]="sliderControl" [auClassName]="'my-0'"></div>
```
quentinderoubaix, line 11: Very very minor... In other examples, we tend to set the string inputs in the format ```auClassName="my-0"``` instead of ```[auClassName]="'my-0'"```

quentinderoubaix:
For better accessibility, the attribute ```aria-orientation``` should be set to ```vertical``` when displaying vertically.

quentinderoubaix:
Another note regarding accessibility, it would be good to put an Accessibility section under the slider examples.

core/lib/services/writables.ts
fbasso: Thanks for your comments. 

For this one, yes, I would prefer to do it separately, to take the time to find all location where it can used (I changed something in the rating I think to use the util `clamp` function instead of a local one, but maybe it can be done differently). Anyway, another PR would be better.

angular/demo/src/app/samples/slider/vertical.route.ts
fbasso: Good catch ! Thanks !

core/lib/slider.ts
```diff
@@ -195,7 +195,7 @@ export function getSliderDefaultConfig() {
 const configValidator: ConfigValidator<SliderProps> = {
 	min: typeNumber,
 	max: typeNumber,
-	stepSize: typeNumber,
+	stepSize: typeNumberInRangeFactory(1, +Infinity),
```
divdavem, line 198: Why is 1 the minimum value? It should be `> 0` so `0.1` should be accepted.

quentinderoubaix (approved):
LGTM!

fbasso:
Thanks for your review @quentinderoubaix !

PR #228, opened 2023-11-14
Slider: aria-label for the handles

fbasso:
Add the aria-label for the handles via a prop.

This PR re-activate the axe tests for the slider (which failed without it).


quentinderoubaix (approved):
LGTM!

fbasso:
Thanks @quentinderoubaix !

PR #230, opened 2023-11-15
feat: support and enforce number and boolean transform functions

quentinderoubaix:


angular/headless/src/lib/coercion.ts
```diff
@@ -0,0 +1,37 @@
...
+ *  ```typescript
+ *  @Input({ transform: auBooleanAttribute }) status: boolean | undefined;
+ *  ```
+ * @param value Value to be transformed.
+ */
+export function auBooleanAttribute(value: unknown): boolean | undefined {
+	if (value === undefined) {
+		return undefined;
+	}
+	return typeof value === 'boolean' || value === undefined ? value : value != null && value !== 'false';
```
divdavem, line 15: Why do you again compare with undefined? It is already done line 12.

quentinderoubaix: Indeed, thanks !

divdavem:
You should also change `reportMissingInputProp` to generate the correct code in case the type is boolean/number.

And you should add the corresponding test cases:
```ts
		{
			code: codeTemplate('', 'myProp: boolean;'),
			errors: [{messageId: 'missingProp', data: {type: 'input', name: 'myProp'}}],
			output: codeTemplate("\n\t@Input({alias: 'auMyProp', transform: auBooleanAttribute}) myProp: boolean | undefined;\n", 'myProp: boolean;'),
		},
		{
			code: codeTemplate('', 'myProp: number;'),
			errors: [{messageId: 'missingProp', data: {type: 'input', name: 'myProp'}}],
			output: codeTemplate("\n\t@Input({alias: 'auMyProp', transform: auNumberAttribute}) myProp: number | undefined;\n", 'myProp: number;'),
		},
```

eslint-plugin/src/angular-check-props.ts
```diff
@@ -320,7 +346,42 @@ export const angularCheckPropsRule = ESLintUtils.RuleCreator.withoutDocs({
 									}
 									const decoratorArguments = (getDecorator(classMember, 'Input')!.expression as TSESTree.CallExpression).arguments;
 									const inputValidAlias = validAlias(name);
-									if (!decoratorArguments.length || decoratorArguments[0].type !== 'Literal' || decoratorArguments[0].value !== inputValidAlias) {
+									const typeAsString = typeToString(inputInfo.type, classMember, context);
+									if (typeAsString === 'boolean | undefined' || typeAsString === 'number | undefined') {
+										const isBoolean = typeAsString === 'boolean | undefined';
```
divdavem, line 351: I would move this up and use `isBoolean` also in the if statement.

eslint-plugin/src/angular-check-props.ts
```diff
@@ -170,16 +170,42 @@ const reportInvalidOutputPropType = (
...
+		node,
+		messageId: type,
+		data: {
+			name,
+			metadata,
+		},
+		*fix(fixer) {
+			const decorator = getDecorator(node, 'Input')!;
+			yield fixer.remove(decorator);
+			yield fixer.insertTextBefore(node, `@Input({alias: '${alias}', transform: ${transform}})`);
```
divdavem, line 194:
```suggestion
			yield fixer.insertTextBefore(node, `@Input(${metadata})`);
```

quentinderoubaix:
thx for the reivew @divdavem ! I have made some changes following your recommendations, when you have time to check ;)

angular/headless/src/lib/coercion.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export function auNumberAttribute(value: unknown): number | undefined {
+	// parseFloat(value) handles most of the cases we're interested in (it treats null, empty string,
+	// and other non-number values as NaN, where Number just uses 0) but it considers the string
+	// '123hello' to be a valid number. Therefore we also check if Number(value) is NaN.
+	if (value === undefined) {
+		return undefined;
+	}
+	const isNumberValue = !isNaN(parseFloat(value as any)) && !isNaN(Number(value));
+	return isNumberValue ? Number(value) : NaN;
```
divdavem, line 36: Is there any reason why we do not do simply `return +value`?

divdavem: Maybe because of null and empty string...

divdavem: So maybe we could simply do:
```ts
return (value === null || value === "") ? NaN : +value;
```

quentinderoubaix: I have to confess I put my brain on the side and copied [angular's implementation](https://github.com/angular/angular/blob/main/packages/core/src/util/coercion.ts)...

I also hesitated to write instead:
```typescript
export function auNumberAttribute(value: unknown): number | undefined {
	if (value === undefined) {
		return undefined;
	}
	return numberAttribute(value);
}
```
What do you think ?

divdavem:
> thx for the reivew @divdavem ! I have made some changes following your recommendations, when you have time to check ;)

Thank you for taking my comments into account!
Apart from the logic to coerce to number, I am fine with the PR.

angular/headless/src/lib/coercion.ts
divdavem: I prefer that indeed.

quentinderoubaix: ok for me !

PR #233, opened 2023-11-16
Slider: don't accept 0 for the step

fbasso:
and add some options to the `typeNumberInRangeFactory` utility.

### DIFF
--- a/core/lib/services/writables.ts
+++ b/core/lib/services/writables.ts
@@ -13,18 +13,39 @@ export const typeNumber: WritableWithDefaultOptions<number> = {
 	normalizeValue: numberNormalizeFn,
 };
 
+export interface TypeNumberInRangeOptions {
+	/** If `true`, the range checking will be strict, excluding the minimum and maximum values. Default is `false`. */
+	strict?: boolean;
+
+	/** If `true`, values outside the range will be clamped to the minimum or maximum. Default is `true`. */
+	useClamp?: boolean;
+}
+
 /**
- * A factory function that creates a type guard function to check and rectify a value is within a specified range.
+ * Factory function for creating a type constraint for numbers within a specified range.
  *
- * @param min - The minimum value allowed.
- * @param max - The maximum value allowed.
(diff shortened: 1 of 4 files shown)


core/lib/services/writables.ts
```diff
@@ -13,18 +13,39 @@ export const typeNumber: WritableWithDefaultOptions<number> = {
 	normalizeValue: numberNormalizeFn,
 };
 
+export interface TypeNumberInRangeOptions {
+	/** If `true`, the range checking will be strict, excluding the minimum and maximum values. Default is `true`. */
```
divdavem, line 17:
```suggestion
	/** If `true`, the range checking will be strict, excluding the minimum and maximum values. Default is `false`. */
```
(cf line 34)

core/lib/services/writables.ts
```diff
@@ -13,18 +13,39 @@ export const typeNumber: WritableWithDefaultOptions<number> = {
...
-export function typeNumberInRangeFactory(min: number, max: number) {
+export function typeNumberInRangeFactory(min: number, max: number, options: TypeNumberInRangeOptions = {}) {
+	const {strict = false, useClamp = true} = options;
 	return <WritableWithDefaultOptions<number>>{
 		normalizeValue: (value) => {
-			const normalizedNumber = numberNormalizeFn(value);
-			return normalizedNumber === INVALID_VALUE ? INVALID_VALUE : clamp(normalizedNumber, max, min);
+			let normalizedNumber = numberNormalizeFn(value);
+			if (normalizedNumber !== INVALID_VALUE) {
+				if (useClamp) {
```
divdavem, line 39: If `strict` is true, it is useless to use clamp:
- either `min < value < max` and `value` does not need to be adjusted (no need to call clamp)
- or `value <= min` and it will be adjusted to `min`
- or `value >= max` and it will be adjusted to `max`
The last two cases are invalid if `strict` is true, so `INVALID_VALUE` will be returned.

divdavem (approved):
Thank you!

PR #235, opened 2023-11-16
fix(slider): Bind min, max and stepSize to clean value

MarkoOleksiyenko:


core/lib/slider.ts
```diff
@@ -389,16 +389,13 @@ export function createSlider(config?: PropsConfig<SliderProps>): SliderWidget {
 	});
 
 	// functions
-	const computeCleanValue = (value: number) => {
-		const min = min$(),
-			max = max$(),
-			stepSize = stepSize$();
+	const computeCleanValue = (value: number, min: number, max: number, stepSize: number) => {
```
divdavem, line 392: As this function no longer use anything from the closure, it could be moved out of it to avoid recreating it each time.

divdavem (approved):
Thank you for your PR! :+1: 
You can merge it after moving the function out of the closure.

PR #236, opened 2023-11-16
feat(slider): Add decimal values support for slider

MarkoOleksiyenko:


core/lib/utils.ts
```diff
@@ -1,2 +1,20 @@
 export const noop = () => {};
 export const identity = <T>(x: T) => x;
+
+/**
+ * @param number - decimal number
+ * @returns the decimal precision of the number
+ */
+export function getDecimalPrecision(number: number) {
+	const regexp = ('' + number).match(/(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/);
```
fbasso, line 9: Just a little point here, but I think it's usually better to create the regexp only once, so define it outside the function. It avoid the recompilation of the regexp pattern.
```
const decimalRegExp = /(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/;
```

then (I would prefer another name for regexp, as it's the result of the match, not the regexp itself)
```
const matches = ('' + number).match(decimalRegExp);
```

core/lib/utils.ts
```diff
@@ -1,2 +1,20 @@
...
+	if (!regexp) {
+		return 0;
+	}
+	return Math.max(
+		0,
+		// Number of digits right of decimal point.
```
fbasso, line 15: I think the code can be simplified with: `(regexp[1]?.length ?? 0)`

MarkoOleksiyenko: and the second one can be (+matches[2] || 0) I believe

fbasso: No, sorry, I re-changed my original code as I saw ts errors locally. The + must be outside the parenthesis as +undefined gives NaN, and the ?? needs undefined to work well

MarkoOleksiyenko: as discussed +Nan || 0 will give 0, so should work fine

core/lib/slider.spec.ts
```diff
@@ -682,6 +682,101 @@ describe(`Slider basic`, () => {
...
+	test(`should correctly increment the handle value when the min is decimal and step is integer`, () => {
+		slider.patch({
+			min: 0.5,
+			max: 3.5,
+			stepSize: 1,
+			values: [0.5],
+		});
+
+		const expectedStateValue = {...defaultStateValues};
+		expectedStateValue.min = 0.5;
```
fbasso, line 695: No need to redefine the values that does not change. 
It's a little bit misleading, as we can't see at a glance what are the impact of an action.

core/lib/slider.spec.ts
```diff
@@ -682,6 +682,101 @@ describe(`Slider basic`, () => {
...
+		expectedStateValue.max = 3.5;
+		expectedStateValue.stepSize = 1;
+		expectedStateValue.values = [0.5];
+		expectedStateValue.sortedValues = [0.5];
+		expectedStateValue.sortedHandles = [{id: 0, value: 0.5, ariaLabel: '0.5'}];
+		expect(state).toStrictEqual(expectedStateValue);
+
+		slider.actions.keydown(keyboardEvent('ArrowUp'), 0);
+
+		expectedStateValue.minValueLabelDisplay = true;
```
fbasso, line 705: Didn't change... ;)

core/lib/slider.ts
```diff
@@ -218,17 +218,20 @@ const configValidator: ConfigValidator<SliderProps> = {
...
 	if (value >= max) {
 		return max;
 	} else if (value <= min) {
 		return min;
 	}
-	const indexMin = Math.floor(value / stepSize);
-	return value % stepSize < stepSize / 2 ? indexMin * stepSize : (indexMin + 1) * stepSize;
+	const indexMin = Math.floor(((value - min) * decimalPower) / intStepSize);
+	return (
+		(((value - min) * decimalPower) % intStepSize < intStepSize / 2 ? indexMin * intStepSize : (indexMin + 1) * intStepSize) / decimalPower + min
```
fbasso, line 233: intStepSize can be factorized

fbasso (commented):
Some comments for the tests.

I had tested with these values:

```
<Slider min={0.33} max={5.55} stepSize={0.2} bind:values={sliderControl} />
```

and the display is wrong (values not rounded).

core/lib/slider.spec.ts
MarkoOleksiyenko: these values change with regard of the default values (that have 0 as min, 100 as maximum and [0] as values), so here it is initial assignment of values for this test case

fbasso (approved):
LGTM ! Thanks for your PR !

PR #243, opened 2023-11-21
feat(slider): Add touch events to slider

MarkoOleksiyenko:
Touch testing is not implemented by Playwright and the unit testing would be not efficient in this case as the methods used are already tested.

### DIFF
--- a/core/src/components/slider/slider.ts
+++ b/core/src/components/slider/slider.ts
@@ -201,6 +201,13 @@ export interface SliderActions {
 	 * @param handleId - numeric id of the handle
 	 */
 	mouseDown(event: MouseEvent, handleId: number): void;
+
+	/**
+	 * Method describing the behavior of the slider handle on touch start event
+	 * @param event - touch event
+	 * @param handleId - number id of the handle
+	 */
+	touchStart(event: TouchEvent, handleId: number): void;
 }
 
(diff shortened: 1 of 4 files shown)


ValentinNelu:
https://github.com/AmadeusITGroup/AgnosUI/assets/105430234/cd35b319-533f-437c-8aa1-8c9ad3f8502b

It feels strange that when we drag, the whole page gets dragged

core/src/components/slider/slider.ts
```diff
@@ -650,6 +657,36 @@ export function createSlider(config?: PropsConfig<SliderProps>): SliderWidget {
 					);
 				}
 			},
+			touchStart(event: TouchEvent, handleId: number) {
+				const handleDrag = (e: TouchEvent) => {
```
divdavem, line 661: As discussed with @ValentinNelu, it is needed to add `e.preventDefault()` here to prevent the scroll of the page from happening at the same time. It is good also to add `event.preventDefault()` at the beginning of `touchStart`. Note that both are correctly done for `mouseDown`.

MarkoOleksiyenko: agree, forgot to add it back when testing it's necessity (y)

ValentinNelu (approved):
Thx for the PR, lgtm

MarkoOleksiyenko:
Thanks for the review @ValentinNelu !

PR #247, opened 2023-11-23
fix eslint plugin error reporting with bindings

divdavem:


divdavem:

Issue #114, opened 2023-08-28
PageObject should we be able to overrides the selectors ?


ExFlo:
Open question.
For people to be able to use the PO with a custom structure slot.

PR #248, opened 2023-11-24
feat: making auUse synchronous

divdavem:
Now that https://github.com/angular/angular/issues/50320 is fixed, it should be possible to change any angular signal at any time and there should be no need anymore to delay the call of directives.


divdavem:
Note that this seems to break opening transitions in the accordion.

quentinderoubaix (approved):
LGTM !
thx for the pr @divdavem

divdavem:

quentinderoubaix:
closes #114

PR #249, opened 2023-11-24
fix(alert): Add alert-dismissible class

MarkoOleksiyenko:


divdavem (approved):
Thank you for this PR! It is fine for me, but maybe you could also, as part of this PR, remove the now useless `ms-auto` class on the close button.

PR #266, opened 2023-11-28
refactor(alert): Extract common part from alert

MarkoOleksiyenko:


core/lib/utils.ts
```diff
@@ -127,3 +127,14 @@ function _traversal<T>(key: string, value: T, fn: TraversalFn, index?: number) {
 export function createTraversal(fn: TraversalFn) {
 	return (json: any) => _traversal('', json, fn);
 }
+/**
+ * Utility method for unit tests
```
divdavem, line 131: This is not used only in unit tests.

PR #267, opened 2023-11-28
feat: speed up stackblitz install

quentinderoubaix:
Stackblitz, for all frameworks, is quite slow to boot / install / serve, and one of the biggest reasons of the slowness is the install step.
This can be sped up by providing a lockfile.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,6 +50,15 @@ jobs:
       - run: npm run build:ci
       - run: npm run format:check
       - run: npm run lint
+      - name: check stackblitz angular lockfile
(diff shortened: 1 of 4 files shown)


divdavem (changes requested):
As discussed, please add in the build github workflow `npm ci --dry-run` for each `package-lock.json` file to check that those new package-lock.json files always match the corresponding package.json.

quentinderoubaix:
> As discussed, please add in the build github workflow `npm ci --dry-run` for each `package-lock.json` file to check that those new package-lock.json files always match the corresponding package.json.

Thanks for the review @divdavem, have updated the PR as discussed.

PR #270, opened 2023-11-29
fix: portal directive should not move element when not needed

divdavem:


quentinderoubaix (approved):
LGTM !

divdavem:

PR #271, opened 2023-11-30
fix: incorrect usage of <slot> in accordion sample

divdavem:


fbasso (approved):
Fine !

divdavem:

PR #280, opened 2023-12-04
fix(demo): keep component active in the menu even when not on "Examples" tab

divdavem:


divdavem:
@quentinderoubaix Thank you for reviewing and merging my PR

PR #288, opened 2023-12-05
ci: make a full install on lockfile maintenance

quentinderoubaix:


quentinderoubaix:

PR #295, opened 2023-12-06
changing structure of core sources

quentinderoubaix:


core/lib/utils/internal/pagination.ts
```diff
@@ -0,0 +1,114 @@
+// Utility functions
```
divdavem, line 1: I think you now have this file twice. This copy should be removed.

PR #307, opened 2023-12-12
feat(toast): Add toast component

MarkoOleksiyenko:


react/demo/src/app/samples/toast/Action.route.tsx
```diff
@@ -0,0 +1,69 @@
...
+const ToastAction = ({widget, state}: ToastContext) => {
+	function actionDemo() {
+		window.alert('Undo');
+	}
+
+	return (
+		<>
+			<div className="d-flex w-100">
+				<div className="d-flex align-items-center flex-grow-1 toast-body">
+					<span className="d-flex me-2" dangerouslySetInnerHTML={{__html: biCheckCircleFill}}></span>
```
quentinderoubaix, line 52: Once https://github.com/AmadeusITGroup/AgnosUI/pull/478 is merged, remove the usage of ```dangerouslySetInnerHTML```

angular/lib/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,221 @@
...
+	 * Callback called when the alert is hidden.
+	 */
+	@Output('auHidden') hidden = new EventEmitter<void>();
+
+	/**
+	 * Callback called when the alert is shown.
+	 */
+	@Output('auShown') shown = new EventEmitter<void>();
+
+	readonly defaultSlots = writable(defaultConfig);
```
quentinderoubaix, line 192: Unfortunately, the type becomes invalid (at build time) if not explicitly set here.
You can reference what was done for the alert.

style-bootstrap/scss/_toast.scss
```diff
@@ -0,0 +1,3 @@
+.au-toast {
+	width: unset;
```
quentinderoubaix, line 2: Set the sass var $toast-max-width instead.

svelte/demo/src/app/samples/toast/Action.route.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
+	import {Toast} from '@agnos-ui/svelte/components/toast';
+	import {createWidgetsDefaultConfig} from '@agnos-ui/svelte/config';
+	import Action from './Action.svelte';
+
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
quentinderoubaix, line 6: This comment can be removed.

svelte/demo/src/app/samples/toast/Action.route.svelte
```diff
@@ -0,0 +1,22 @@
...
+	const widgetsConfig$ = createWidgetsDefaultConfig();
+	$widgetsConfig$.toast = {...$widgetsConfig$.toast, autohide: false, dismissible: true, className: 'text-bg-success', slotStructure: Action};
```
quentinderoubaix, line 8: You can replace this with:
```typescript
createWidgetsDefaultConfig().set({toast: {autohide: false, dismissible: true, className: 'text-bg-success', slotStructure: Action}});
```

svelte/demo/src/app/samples/toast/Action.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+	import {Slot} from '@agnos-ui/svelte/slot';
+	import type {ToastSlots} from '@agnos-ui/svelte/components/toast';
+	import biCheckCircleFill from 'bootstrap-icons/icons/check-circle-fill.svg?raw';
+	import biArrowCounterClockwise from 'bootstrap-icons/icons/arrow-counterclockwise.svg?raw';
+
+	// cf https://github.com/ota-meshi/eslint-plugin-svelte/issues/348
+	type $$Props = ToastSlots['structure']; // eslint-disable-line @typescript-eslint/no-unused-vars
```
quentinderoubaix, line 8: You can remove the comments.

svelte/demo/src/app/samples/toast/Action.svelte
```diff
@@ -0,0 +1,39 @@
...
+	export let widget: $$Props['widget'];
+
+	function actionDemo() {
+		window.alert('Undo');
+	}
+</script>
+
+<div class="d-flex w-100">
+	<div class="d-flex align-items-center flex-grow-1 toast-body">
+		<!-- eslint-disable-next-line svelte/no-at-html-tags -->
```
quentinderoubaix, line 21: Once https://github.com/AmadeusITGroup/AgnosUI/pull/478 is merged, you can remove the eslint-disable comments.

svelte/demo/src/app/samples/toast/Default.route.svelte
```diff
@@ -0,0 +1,37 @@
...
+	function reset() {
+		visible = true;
+	}
+</script>
+
+<p class="mb-2">
+	Color schemes are based on the Bootstrap classes, in order to have out-of-the-box text accessibility use the helper classes <code>.text-bg-*</code>,
+	e.g. <code>.text.bg-primary</code>
+</p>
+<Toast className="text-bg-primary" slotHeader="I am header" autohide={false} dismissible={true} bind:visible>Simple primary toast</Toast>
```
quentinderoubaix, line 19: You can just put ```dismissible``` instead of ```dismissible={true}```

svelte/demo/src/app/samples/toast/Default.route.svelte
```diff
@@ -0,0 +1,37 @@
+<script lang="ts">
+	import {Toast} from '@agnos-ui/svelte/components/toast';
+	import {createWidgetsDefaultConfig} from '@agnos-ui/svelte/config';
+
+	const widgetsConfig$ = createWidgetsDefaultConfig();
+	$widgetsConfig$ = {toast: {autohide: false, dismissible: false}};
```
quentinderoubaix, line 6: Config can be one line here too.

svelte/demo/src/app/samples/toast/Dynamic.route.svelte
```diff
@@ -0,0 +1,76 @@
...
+	const positions = Object.entries(ToastPositions).map((entry) => {
+		return {
+			value: entry[1],
+			label: entry[0],
+		};
+	});
+	let position = ToastPositions.topLeft;
+
+	function createToasts() {
+		const {subscribe, update} = writable(new Map(Object.entries(ToastPositions).map((entry) => [entry[1] as string, [] as Partial<ToastProps>[]])));
```
quentinderoubaix, line 27: Object.values

quentinderoubaix (commented):
Thx for the work @MarkoOleksiyenko !

quentinderoubaix:
As discussed, please check the display in Mobile too, as the dynamic sample has horizontal scrolling + the action toast is squashed..

angular/demo/src/app/samples/toast/default.route.ts
```diff
@@ -0,0 +1,45 @@
+import type {ToastComponent} from '@agnos-ui/angular';
+import {AgnosUIAngularModule, provideWidgetsConfig} from '@agnos-ui/angular';
+import {Component, ViewChild} from '@angular/core';
+
+@Component({
+	standalone: true,
+	imports: [AgnosUIAngularModule],
+	providers: [
+		provideWidgetsConfig((config) => {
+			config.toast = {...config.toast, autohide: false, dismissible: false};
```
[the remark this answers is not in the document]
MarkoOleksiyenko: feels cleaner to do that it in the provideWIdgetsConfig as we have everything in one place and not scattered between providers and the class

core/src/components/toast/toast.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+	});
+
+	test(`should close on method call`, () => {
+		const expectedState = state;
+		expect(expectedState.visible).toBe(true);
+		toast.api.close();
+		expectedState.visible = false;
+		expectedState.hidden = true;
+		expect(state).toEqual(expectedState);
+	});
```
fbasso, line 40: You can use the `assign` helper for a more concise code :

```typescript
import {assign} from '../../../../common/utils';
...
expect(state).toEqual(assign(expectedState, {
    visible: false,
    hidden: true,
}));
```

core/src/components/toast/toast.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+				onHidden() {
+					promiseOnHidden.resolve();
+					onHiddenCounter++;
+					promiseOnHidden = promiseWithResolve();
+				},
+				onVisibleChange(event) {
+					onVisibleChangeCounter++;
+				},
+				delay: 2000,
+			},
```
fbasso, line 109: Maybe something shorter is enough (100) ?

core/src/components/toast/toast.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+		const element = document.createElement('div');
+		element.innerHTML = '<div>body</div>';
+		const toastEvents = createToast({
+			props: {
+				onHidden() {
+					promiseOnHidden.resolve();
+					onHiddenCounter++;
+					promiseOnHidden = promiseWithResolve();
+				},
+				onVisibleChange(event) {
```
fbasso, line 105: event is not used, it can be removed

core/src/components/toast/toast.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+					promiseOnShown.resolve();
+					onShownCounter++;
+					promiseOnShown = promiseWithResolve();
+				},
+				onHidden() {
+					promiseOnHidden.resolve();
+					onHiddenCounter++;
+					promiseOnHidden = promiseWithResolve();
+				},
+				onVisibleChange(event) {
```
fbasso, line 72: event is not used, it can be removed

e2e/demo-po/toast.po.ts
```diff
@@ -0,0 +1,26 @@
...
+
+	get locatorAddToastButton(): Locator {
+		return this.locatorRoot.locator(this.locators.addToastButton);
+	}
+
+	get locatorPositionSelect(): Locator {
+		return this.locatorRoot.locator(this.locators.positionSelect);
+	}
+
+	get locatorToastContainer(): Locator {
```
fbasso, line 23: This return all the containers. In the associated tests, it seems more useful to target a specific container.

In this case, it would be better to write it as a function instead of a getter:

```typsescript
    locatorToastContainer(index: number): Locator {
        return this.locatorRoot.locator(this.locators.toastContainer).nth(index);
	}
}

```

```typsescript
// Now you can use :
await toastDemoPO.locatorToastContainer(4);

instead of
(await toastDemoPO.locatorToastContainer.all()).at(4);
```

page-objects/lib/toast.po.ts
```diff
@@ -0,0 +1,19 @@
...
+export class ToastPO extends BasePO {
+	selectors = structuredClone(toastSelectors);
+
+	override getComponentSelector(): string {
+		return this.selectors.rootComponent;
+	}
+
+	get locatorCloseButton(): Locator {
+		return this.locatorRoot.locator(this.selectors.closeButton);
+	}
```
fbasso, line 18: I think there is a missing state() method, to return the displayed state of the widget. All revelant information, like in the rating.po for example.

For example, the state could return:

- The class applied,
- The close button (undefined if no button)
- The aria-label for the close button (undefined if no button)
- The header text (undefined if no header)
- The body text

The issue is that the toast (this is the same for the alert, at the time this architecture was not in place) is not well tested form an e2e point of view (we don't test that everything is "connected").
The current e2e test is fine, but it tests only the containers, not the toasts themselves.

To test the toast with different configurations, no need to add samples, you can use the playground. Even if it doesn't work fully, it allows to play with the props easily. 

For example, you can check the returned state for 

`/components/toast/playground#{"props":{"slotDefault":"This is a toast","autohide":false,"className":"text-bg-primary"}}`, 

then change the hash for 

`components/toast/playground#{"props":{"slotDefault":"This is a toast","autohide":false,"className":"text-bg-primary","slotHeader":"header"}}`

And check the changes compared to the expected state. No need to check deeply (this is the UT purposes), but at least go through all the concept once to check that everything is "plugged".

page-objects/lib/toast.po.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	async state() {
+		return await this.locatorRoot.evaluate((rootNode: HTMLElement, selectors) => {
+			const body = rootNode.querySelector<HTMLDivElement>(selectors.body)?.innerText;
+			const header = rootNode.querySelector<HTMLDivElement>(selectors.header)?.innerText;
+			return {
+				rootClasses: rootNode.className.trim().split(' ').sort(),
+				body,
+				header,
```
fbasso, line 29: You can also add if th close button is here, its aria-label attribute I think

e2e/toast/toast.e2e-spec.ts
```diff
@@ -0,0 +1,65 @@
...
+		const toastPO = new ToastPO(page);
+
+		const expectedState: State = {
+			rootClasses: ['au-toast', 'show', 'text-bg-primary', 'toast'],
+			body: 'This is a toast',
+			header: 'Header',
+		};
+
+		await expect.poll(() => toastPO.state()).toEqual(expectedState);
+		await expect(toastPO.locatorCloseButton).toHaveCount(0);
```
fbasso, line 63: Maybe you can continue the test by changing the dismissible and check that teh close button is here

page-objects/lib/toast.po.ts
MarkoOleksiyenko: makes sense to add it in the state, as we have the button as part of the component

e2e/toast/toast.e2e-spec.ts
MarkoOleksiyenko: not sure that the test should modify the dismissible on the fly as it is not an expected behavior of the component

MarkoOleksiyenko: no, i mean that i think it is susfficient to test whether it is there or not in different tests as users are not supposed to remove/add button on the fly

react/demo/src/app/samples/toast/Action.route.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export default ActionToastDemo;
+
+const ToastAction = ({widget, state}: ToastContext) => (
+	<div className="d-flex w-100">
+		<div className="d-flex align-items-center flex-grow-1 toast-body">
+			<span className="d-flex me-2">{BiCheckCircleFill({})}</span>
+			<Slot slotContent={state.slotDefault} props={{widget, state}}></Slot>
+			<button type="button" className="btn btn-sm ms-auto text-bg-success" onClick={() => window.alert('Undo')}>
+				<span className="me-2" dangerouslySetInnerHTML={{__html: biArrowCounterClockwise}}></span>Undo
```
quentinderoubaix, line 49: dangerouslySetInnerHTML to remove

quentinderoubaix (approved):
LGTM !

e2e/toast/toast.e2e-spec.ts
fbasso: Yes, you're right ;)

fbasso (approved):
Fibne for me, thansk @MarkoOleksiyenko !

PR #319, opened 2023-12-14
fix: generation of exports on windows

quentinderoubaix:


divdavem:
Should we have a build also on a Windows machine in GitHub CI?

fbasso (approved):
LGTM !

PR #322, opened 2023-12-15
[WIP] refactor: removing bindableDerived

divdavem:


divdavem:
Several usages of `bindableDerived` have been removed in other PRs (cf #324 and #334)
We should still get rid of the remaining `bindableDerived` usages in baseTransitions.ts and in select.ts.
I am closing this PR because it is based on very old code and does not contain a working replacement for those remaining usages.

PR #323, opened 2023-12-15
chore: sub exports for svelte

quentinderoubaix:


svelte/headless/package.json
```diff
@@ -10,12 +10,41 @@
 		".": {
 			"types": "./dist/index.d.ts",
 			"svelte": "./dist/index.js"
+		},
+		"./components/*": {
+			"types": "./dist/components/*/*.d.ts",
+			"default": "./dist/components/*/*.js"
```
divdavem, line 16: I think we should be consistent.
Should we put "svelte" or "default" everywhere?

divdavem (approved):
Thank you for your PR! :+1:

quentinderoubaix:

PR #332, opened 2023-12-18
refactor: directly calls widget factory with initial props in react

divdavem:


MarkoOleksiyenko (approved):

fbasso:
Maybe we should add this feature in other frameworks too, to see if something common can be done (so it maybe change the code of this PR).

The point is to check that no onXXXChange happens at initialization time, which is incorrect, and which probably happen now in all frameworks

What do you think ?

divdavem:
> Maybe we should add this feature in other frameworks too, to see if something common can be done (so it maybe change the code of this PR).
I have opened PR https://github.com/AmadeusITGroup/AgnosUI/pull/342 for Angular.
I will open another PR for Svelte.

divdavem:
@MarkoOleksiyenko @fbasso Thank you for your review and comment.

PR #339, opened 2023-12-20
chore: removing svelte headless tsconfig paths

quentinderoubaix:


divdavem (approved):
It is a shame that `@sveltejs/package` (or `typescript` ?) does not automatically handle those types correctly.
Thank you for investigating and working around the issue!

quentinderoubaix:

PR #345, opened 2023-12-22
feat(slider): Add RTL support

MarkoOleksiyenko:


style-bootstrap/scss/_slider.scss
```diff
@@ -130,10 +130,18 @@ $prefix: au-;
 			left: var(--#{$prefix}slider-label-min-left);
 		}
 
+		&.au-slider-label-min-rtl {
+			right: var(--#{$prefix}slider-label-max-right);
```
MarkoOleksiyenko, line 134: rename variables and check out the mixins

core/src/components/slider/slider.ts
```diff
@@ -83,6 +92,11 @@ export interface SliderCommonPropsAndState extends WidgetsCommonPropsAndState {
 	 * If `true` the min and max labels are displayed on the slider
 	 */
 	showMinMaxLabels: boolean;
+
+	/**
+	 * It `true` slider display is inversed
+	 */
+	rightToLeft: boolean;
```
ExFlo, line 99: Hi use RTL everywhere instead of rightToLeft. Thx

demo/src/routes/[framework]/components/slider/examples/+page.svelte
```diff
@@ -15,6 +16,9 @@
 <Section label="Vertical slider" id="range" level={2}>
 	<Sample title="Vertical slider" sample={verticalSample} height={627} />
 </Section>
+<Section label="Right to left slider" id="range" level={2}>
```
Comment (ExFlo) [inline demo/src/routes/[framework]/components/slider/examples/+page.svelte:19]: id is not right it has to be adjusted to the new ex

core/src/components/slider/slider.ts
```diff
@@ -250,6 +266,41 @@ const computeCleanValue = (value: number, min: number, max: number, intStepSize:
...
+ */
+const updateDirtyValue = (
+	handleIndex: number,
+	_dirtyValues$: WritableSignal<number[]>,
+	values: number[],
+	stepSize: number,
+	updateDirection: number,
+) => {
+	_dirtyValues$.update((value) => {
+		value = [...value];
```
ValentinNelu, line 285: why don't we just define a new const instead of reusing the same name? I think it would be more readable

core/src/components/slider/slider.ts
```diff
@@ -383,7 +445,9 @@ export function createSlider(config?: PropsConfig<SliderProps>): SliderWidget {
 			return true;
 		}
 		const maxLabelWidth = maxLabelWidth$();
-		return !valuesPercent$().some((percent) => percent > 100 - maxLabelWidth - 1);
+		return rtl$()
+			? !valuesPercent$().some((percent) => 100 - percent < maxLabelWidth + 1)
+			: !valuesPercent$().some((percent) => percent > 100 - maxLabelWidth - 1);
 	});
 	// TODO define the intersection value
```
ValentinNelu, line 452: just for curiosity, do we still need the TODO?

MarkoOleksiyenko: yep, haven't actually tested with the huge numbers, will come with user testing

core/src/components/slider/slider.ts
MarkoOleksiyenko: you mean _dirtyValue$? as it is still a writable signal it makes sense to keep the old notation as it is in the same core file and user knows to what it pertains

ValentinNelu: I meant const newValue = [...value]; and so on

ValentinNelu (approved):

ExFlo (approved):

PR #349, opened 2023-12-28
core: enable linter on Promise

fbasso:
The PR does not pass yet, this is just to discuss about it first.

It enables two rules:

  - https://typescript-eslint.io/rules/no-floating-promises/
  - https://typescript-eslint.io/rules/no-misused-promises/

With this PR, promise need to be explicitly handled. For example:

```typescript
  await getAsyncFn();

  // or, if we don't want to wait:
  void getAsyncFn();
```

The point is to avoid issues/mistakes in test as [here](https://github.com/AmadeusITGroup/AgnosUI/blob/ad8da0864a080db682d78e1220f9c37deadd348b/e2e/floatingUI/floatingUI.e2e-spec.ts#L13), where ```toBeVisible()``` is not awaited.

But probably having this rule enabled everywhere is safer.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,6 +25,8 @@
 	],
 	"rules": {
 		"prefer-const": ["error", {"destructuring": "all"}],
+		"@typescript-eslint/no-floating-promises": "error",
+		"@typescript-eslint/no-misused-promises": ["error", {"checksVoidReturn": false}],
 		"@typescript-eslint/no-unused-vars": ["error", {"vars": "all", "args": "none", "ignoreRestSiblings": false}],
 		"@typescript-eslint/no-non-null-assertion": "off",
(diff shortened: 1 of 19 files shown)


fbasso:
I've changed the code accordingly, to see the required changes.

I just wonder if it's fine or not to have `void promise`in some demo.

quentinderoubaix (approved):
LGTM!

Thanks the PR, it will be clear how we use our promises and we will avoid making the silly mistake to forget an `await`.

fbasso:
Thanks @quentinderoubaix ! I think it's clearer too. I will just wait for @divdavem,to se if he has some comments/ideas about it.

divdavem (approved):

fbasso:
Thanks for your reviews @divdavem  and @quentinderoubaix !

PR #351, opened 2024-01-03
fix: initial 500 error when navigating from home page

quentinderoubaix:
Launching the Vite server in local would result in a temporary 500 error navigating from the home page to any other page.
This is due to the fact some dependencies (highlight.js, floating-ui and tansu) are needed everywhere but the home page, thus vite would not optimize the dependencies at initial startup.

This PR proposes to list those dependencies as "to always optimize".


fbasso (approved):
Fine for me !

quentinderoubaix:
thx for the review @fbasso !

PR #353, opened 2024-01-03
feat: setup wireit

quentinderoubaix:


angular/headless/package.json
```diff
@@ -12,14 +12,46 @@
...
+				"../../core:build"
+			],
+			"files": [
+				"src/**",
+				"!src/generated",
+				"tsconfig.json",
+				"angular.json"
+			],
+			"output": [
+				"dist/**"
```
divdavem, line 37: Should you add `src/generated` in the output?

divdavem:
Thank you for this PR. wireit looks very interesting.
I am thinking that if a task is managed by wireit (especially for caching), we should always pass through wireit to execute it to best benefit from the cache. That's not the case currently, especially for `generate:exports` that is run right from the start (in the prepare script). This means we should avoid having "npm run ..." in wireit commands (because that means we can do the same `npm run ...` without passing through wireit).
What do you think? Maybe we should integrate this PR and then improve that later?

Also, [this build](https://github.com/AmadeusITGroup/AgnosUI/actions/runs/7462755003/attempts/1) worries me a bit because it failed, and re-running it succeeded. Did you investigate why? Is there any race condition when wireit runs? Does wireit run things in parallel and in some scenario, could it lead to some errors? I find this strange.

quentinderoubaix:
> Thank you for this PR. wireit looks very interesting. I am thinking that if a task is managed by wireit (especially for caching), we should always pass through wireit to execute it to best benefit from the cache. That's not the case currently, especially for `generate:exports` that is run right from the start (in the prepare script). This means we should avoid having "npm run ..." in wireit commands (because that means we can do the same `npm run ...` without passing through wireit). What do you think? Maybe we should integrate this PR and then improve that later?
> 
> Also, [this build](https://github.com/AmadeusITGroup/AgnosUI/actions/runs/7462755003/attempts/1) worries me a bit because it failed, and re-running it succeeded. Did you investigate why? Is there any race condition when wireit runs? Does wireit run things in parallel and in some scenario, could it lead to some errors? I find this strange.

thx for our initial review @divdavem !

I have updated the PR to wire wireit everywhere that seemed interesting.
Investigating the build that failed, it seemed linked to how the cache is managed with the wire github action.
For the moment, I propose to merge this PR without enabling the github cache, and we would enable it with another PR.

What do you think ?

angular/demo/package.json
```diff
@@ -4,13 +4,54 @@
...
+		"dev:coverage": "wireit",
+		"build": "wireit",
+		"build:coverage": "wireit",
 		"watch": "ng build --watch --configuration development",
 		"preview": "node scripts/preview.cjs dist --port 4200 --single"
 	},
+	"wireit": {
+		"dev": {
+			"command": "ng serve",
+			"dependencies": [
```
divdavem, line 17: As discussed, let's add a dependency on build (because of the issue with Angular).

divdavem (approved):
Thank you for your PR! I think it can finally be merged now.

quentinderoubaix:
thx for your review @divdavem !

PR #359, opened 2024-01-10
test: re-enable sampleMarkup

fbasso:
It has been skipped by mistake since [the big files restructuring](https://github.com/AmadeusITGroup/AgnosUI/commit/4e1fe9fc14255a2fd694e1b9de6419893c4709e1).

I had to redo some snapshots (code review required !), but it seems there is a discrepancy between frameworks with the `slider/right-to-left` check.


playwright.config.ts
```diff
@@ -45,10 +45,23 @@ const projects: PlaywrightTestConfig<FixtureOptions>['projects'] = includesDemo
...
+	projects.push({
+		name: `${framework.name}:samplesMarkup`,
+		testMatch: 'samplesMarkup.e2e-spec.ts',
+		use: {
+			...devices['Desktop Chrome'],
+			framework: framework.name,
+			baseURL: framework.url,
+		},
+	});
+});
```
divdavem, line 58: Why do you want to run this test only on Chrome?

quentinderoubaix (approved):
LGTM !

PR #360, opened 2024-01-10
feat: move the framework selection into the side menu

andreascorti:
In this way, it is easier to understand that the framework selected is global for your documentation


divdavem (approved):
Thank you for your PR!

PR #361, opened 2024-01-11
docs: focus track MD

andreascorti:
Documentation of the focus track service.
The idea would be to keep a docs folder for all the generic markdown and then include them in the demo application whenever needed.


andreascorti:
Already merged

PR #365, opened 2024-01-16
feat: markdown doc

quentinderoubaix:


andreascorti:
Thanks for the MD integration @quentinderoubaix. 
As discussed, I think we should have all MDs at root level to make it easy to have a look to the documentation directly from IDE / github (in a second step with a slug convention).

demo/src/routes/docs/[framework]/getting-started/installation/installation.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Create a new project
+
+If you haven't already, create a new project where you want to use AgnosUI. You can use the `blabla` for this.
+
+## Install AgnosUI
+
+AgnosUI provides two flavours of the widgets for you to choose from:
+
+- **Headless**: The {$selectedFramework$}-headless package contains the headless functionality of the widget library. It can be used with any front-end CSS framework. It uses the core of AgnosUI which is framework agnostic.
```
Comment (andreascorti) [inline demo/src/routes/docs/[framework]/getting-started/installation/installation.md:17]: If we use variables in the MD like `{$selectedFramework$}`, I think you will need a custom renderer else it will be rendered as html

demo/src/routes/docs/[framework]/[...slug]/+page.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+	import Markdown from '$lib/markdown/Markdown.svelte';
+	export let data;
+</script>
+
+<svelte:head>
+	<meta name="description" content="Focus track service documentation" />
```
Comment (andreascorti) [inline demo/src/routes/docs/[framework]/[...slug]/+page.svelte:7]: if you want to keep description, it should be a param

andreascorti (approved):
LGTM

quentinderoubaix:

PR #367, opened 2024-01-16
feat: default tsconfig with alias, allowing to go to declaration in VSCode

quentinderoubaix:


angular/demo/tsconfig.json
```diff
@@ -3,7 +3,14 @@
 	"extends": "../tsconfig.json",
```
divdavem, line 3: Should'nt all `tsconfig.json` extend from their corresponding `tsconfig.build.json` and only change the paths as it is done in angular/headless?

quentinderoubaix: thx for the comment @divdavem !
have updated the PR accordingly

divdavem (approved):

PR #379, opened 2024-01-19
fix: eslint not working in vscode for some files in Angular

divdavem:


quentinderoubaix (approved):
LGTM !
Thanks for the fix 👍

PR #390, opened 2024-01-22
fix(responsiveness): Main content / menu sizes

### LINKED ISSUE #388: Responsiveness SideMenu and TOC
**Describe the bug**
There is not a max-width for left and right menu, or text-ellipses. 
If a text is quite long, SideMenu or TOC will take space to the main content, which should not be the correct behaviour.

**To Reproduce**
Compare a Documentation page:
image
to a blog one:
image

You can already see the difference. Try to modify the labels for SideMenu and TOC and you will also notice the issue.

**Expected behavior**
SideMenu and TOC should have a max-width, so that it does not narrow too much the main content.

andreascorti:
closes #388

### DIFF
--- a/demo/src/lib/markdown/renderers/MdImage.svelte
+++ b/demo/src/lib/markdown/renderers/MdImage.svelte
@@ -10,4 +10,4 @@
 	$: void getSrc(href);
 </script>
 
-<img {src} {title} alt={text} />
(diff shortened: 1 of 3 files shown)


quentinderoubaix (approved):
LGTM !

PR #391, opened 2024-01-22
refactor: move to the new (non-experimental) API for decorators

divdavem:


divdavem:
This PR is no longer relevant, `CachedProperty` was removed in #725.

PR #397, opened 2024-01-22
feat: add previous and next link to each doc page

quentinderoubaix:


demo/src/lib/markdown/renderers/MdLink.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+	export let href = '';
+	export let title: string;
+
+	const validMdRegex = /^\d{2}-[a-zA-Z-]*\.md$/g;
```
andreascorti, line 5: This won't work if you reference another category.
Example: `(Focus Track)[../services/01-FocusTrack.md)`

Also, the extension is not mandatory in most of the markdown renderer; you can have both URL path and file path

andreascorti (commented):
Thanks @quentinderoubaix for the PR. 
I have a global question: you implemented the navigation only to a single section, and it applies only to the docs with slug. Two remarks:
1. Shouldn't it be also on components?
2. Shouldn't you be able to jump to one section to another with the next / prev links? (i.e When in "Extending", I would expect to navigate next to "Accordion".

My suggestions could be into a new issue

demo/src/lib/markdown/renderers/MdLink.svelte
quentinderoubaix: Thx for the suggestion, have updated the MdLink renderer to support relative markdown links

quentinderoubaix:
> Thanks @quentinderoubaix for the PR. I have a global question: you implemented the navigation only to a single section, and it applies only to the docs with slug. Two remarks:
> 
> 1. Shouldn't it be also on components?
> 2. Shouldn't you be able to jump to one section to another with the next / prev links? (i.e When in "Extending", I would expect to navigate next to "Accordion".
> 
> My suggestions could be into a new issue

Thx for the review @andreascorti !
I have updated the PR to support multiple sections, while also trying to separate a bit better.

andreascorti (approved):
LGTM!

quentinderoubaix:

PR #399, opened 2024-01-23
feat(menu): Add collapsible sections to side menu

### LINKED ISSUE #389: Make the menu sections collapsilble
**Description**
We start to have many sections / sub routes in our Documentation, and it will grow bigger the more components we include.
Let's make the sections collapsible (collapsed by default).

MarkoOleksiyenko:
closes #389

### DIFF
--- a/demo/src/routes/menu/CollapsibleSection.svelte
+++ b/demo/src/routes/menu/CollapsibleSection.svelte
@@ -1,0 +1,61 @@
+<script lang="ts">
+	import {collapseVerticalTransition} from '@agnos-ui/core/services/transitions/bootstrap';
+	import {createTransition} from '@agnos-ui/core/services/transitions/baseTransitions';
+	import {writable} from '@amadeus-it-group/tansu';
+
+	export let headerText: string;
+	export let defaultVisible: boolean;
+
+	const paramTransition$ = writable(collapseVerticalTransition);
+	const paramAnimation$ = writable(true);
+	const paramVisible$ = writable(defaultVisible);
+
+	const {
+		stores: {visible$},
+		api: {toggle},
+		directives: {directive},
+	} = createTransition({
+		props: {
+			animation: paramAnimation$,
+			visible: paramVisible$,
+		},
+	});
+</script>
+
+<div class="mb-2">
+	<button
+		class="btn p-1 d-flex align-items-center btn-header fw-semibold w-100 justify-content-between"
+		aria-expanded={$visible$}
+		on:click={() => toggle()}
+		>{headerText}
+		<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 100 100" class="ms-1">
+			<rect class="horizontal" x="20" y="45" width="60" height="10" fill="black" />
+			<rect class="vertical" class:expanded={$visible$} x="45" y="20" width="10" height="60" fill="black" />
+		</svg>
+	</button>
+	<div class="contents" class:collapse={!defaultVisible} use:directive={{transition: $paramTransition$, animation: $paramAnimation$}}>
+		<slot />
+	</div>
+</div>
+
+<style lang="scss">
+	.btn-header:focus,
+	.btn-header:hover,
+	.btn-header:focus:active {
+		background-color: var(--bs-secondary-bg);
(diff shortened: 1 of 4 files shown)


demo/src/routes/menu/CollapsibleSection.svelte
```diff
@@ -0,0 +1,50 @@
...
+
+<div class="mb-2">
+	<button class="btn btn-light btn-lg p-1 d-flex align-items-center btn-header" aria-expanded={$visible$} on:click={() => toggle()}
+		>{headerText}
+		<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 100 100">
+			<rect class="horizontal" x="20" y="40" width="60" height="20" fill="black" />
+			<rect class="vertical" class:expanded={$visible$} x="40" y="20" width="20" height="60" fill="black" />
+		</svg>
+	</button>
+	<div class="contents" use:directive={{transition: $paramTransition$, animation: $paramAnimation$}}>
```
MarkoOleksiyenko, line 31: wanted to remove the content with an if-block when it is hidden, however, then the sveltekit prerendering is not working automatically, so maybe not worth for a list of links WDYT?

demo/src/routes/menu/CollapsibleSection.svelte
```diff
@@ -0,0 +1,50 @@
+<script>
+	import {collapseVerticalTransition} from '@agnos-ui/core/services/transitions/bootstrap';
+	import {createTransition} from '@agnos-ui/core/services/transitions/baseTransitions';
+	import {writable} from '@amadeus-it-group/tansu';
+
+	export let headerText;
+	const paramTransition$ = writable(collapseVerticalTransition);
+	const paramAnimation$ = writable(true);
+	const paramVisible$ = writable(false);
```
andreascorti, line 9: shouldn't be open by default?

demo/src/routes/menu/CollapsibleSection.svelte
```diff
@@ -0,0 +1,50 @@
...
+	} = createTransition({
+		props: {
+			animation: paramAnimation$,
+			visible: paramVisible$,
+		},
+	});
+</script>
+
+<div class="mb-2">
+	<button class="btn btn-light btn-lg p-1 d-flex align-items-center btn-header" aria-expanded={$visible$} on:click={() => toggle()}
```
andreascorti, line 24: I would probably add `justify-content-between` and `w-100` here to give more order in the layout

demo/src/routes/menu/CollapsibleSection.svelte
```diff
@@ -0,0 +1,50 @@
...
+		>{headerText}
+		<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 100 100">
```
andreascorti, line 26: shouldn't we reuse icons of accordion?

demo/src/routes/menu/CollapsibleSection.svelte
andreascorti: I do agree with you, moreover I suggested to open menus by default, which it means you load all the menu at page load.

demo/src/routes/menu/CollapsibleSection.svelte
MarkoOleksiyenko: we decided to open only the currently visible section by default (e.g. Getting started, when the Introduction is open), so that the space is preserved

demo/src/routes/menu/CollapsibleSection.svelte
MarkoOleksiyenko: i like it, good suggestion!

demo/src/routes/menu/CollapsibleSection.svelte
MarkoOleksiyenko: wanted to do something different from accordion, this feels different and more fun :) FYI didn't use the accordion itself cause the only massive thing that connects these two is a transition and this is a pretty simple case

andreascorti (approved):
LGTM thanks @MarkoOleksiyenko !

quentinderoubaix (approved):
LGTM !

PR #407, opened 2024-01-24
fix: removing loading="lazy" on demo samples iframe (as Firefox bugs with it)

divdavem:


fbasso (approved):
Fine for me, thanks @divdavem !

divdavem:

PR #408, opened 2024-01-24
demo: service worker downloads assets on request rather than when installed

divdavem:


quentinderoubaix (approved):
LGTM !

divdavem:

PR #410, opened 2024-01-24
docs: Directives

### LINKED ISSUE #401: [Doc] Directives
**Implementation**
Explain the concept of directives, how we create directives and use them in the different frameworks.

andreascorti:
closes #401

### DIFF
--- a/docs/01-Headless/03-Directives.md
+++ b/docs/01-Headless/03-Directives.md
@@ -1,5 +1,156 @@
 # Directives
 
-## TODO
+## Overview
 
-To be done, issue #401
+Directives in AgnosUI are inspired by the **actions** in Svelte. These directives are element-level lifecycle functions that are executed when the element is created. They are updated (if an update function is provided) when a parameter changes, and finally, they are executed when the element is destroyed.
+These directives are typically used to add custom event handlers. For instance, core services like [focus track](../04-Services/01-Focustrack.md) and [floating ui](../04-Services/03-Floating-UI.md) create directives that are utilized by components such as [Select](../components/select/examples).
+
+## Usage
+
+The usage of this function is dependent on the JavaScript framework (if any) that the application utilizes. For instance, in this guide, we are using simple TypeScript without any specific framework.
+
+### How to create a directive
+
+A directive is a function that takes as input an `HTMLElement` and an optional parameter, for example an object representing a configuration.
+This function will execute a first time when the directive is applied to an element (usually when the DOM element is created) and either it returns `void`, or 2 optional functions:
+
+- `update` function, called when the context changed and you need either to update the directive configuration or to re-execute partially the original function or modify some internal state of the directive.
+- `destroy` function, used to remove eventual listeners, subscriptions or any reactive objects created by the directive. This function is usually called when the DOM element on which you applied the directive gets removed from the DOM.
+
+```typescript
+const createSampleDirective: Directive<string> = (element: HTMLElement, text: string) => {
+	console.log('Directive has been executed on node ', element.id);
+	const clickListener = (event: Event) => {
+		console.log(text, event.target);
+	};
+	element.addEventListener('click', clickListener);
+	return {
+		update: (newText) => (text = newText),
+		destroy: () => {
+			element.removeEventListener('click', clickListener);
+			console.log('Destroy function executed ', element.id);
+		},
+	};
+};
+```
+
+This example shows a very simple directive that creates an event listener on `click` events that happen on the HTML element and then print the parameter `text` in the console, along with the clicked target.
+
+Both `update` and `destroy` functions are provided by the directive, as following:
+
+- `update`: gives a way to modify the original text to something new, when the context changes (see in the next section).
+- `destroy`: remove the previously created listener, so that you don't bloat the main thread with unused listeners.
+
+### How to use a directive
+
+Considering this HTML page, in which you have a `div` container `focus-element` that includes 2 `button`, an input text `clickText` and another `button` that removes the container.
+
+```html
+<div id="content">
+	<div id="focus-element">
+		<button>button 1</button>
+		<button>button 2</button>
+	</div>
+	<hr />
+	<input id="clickText" type="text" />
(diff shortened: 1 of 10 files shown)


docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,120 @@
...
+observer.observe(document.getElementById('content'), {childList: true});
+```
+
+### Angular usage
+
+//TODO show the utility provided by agnos (2 directives auUse auUseParams)
+
+### React usage
+
+//TODO ref usage -> utility provided by agnos to plug
```
fbasso, line 108: As a reminder, I would say : `utilities` -> there is also the same use case as in Angular, several directives used on the same element need to be merged.

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,120 @@
...
+const createMyDirective: Directive<string> = (element: HTMLElement, text: string) => {
+	console.log('Directive has been executed on node ', element.id);
+	const clickListener = (event: Event) => {
+		console.log(text, event.target);
+	};
+	element.addEventListener('click', clickListener);
+	return {
+		update: (newText) => (text = newText),
+		destroy: () => {
+			node.removeEventListener('click', clickListener);
```
fbasso, line 30: node -> element

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,120 @@
...
+focus-element click
+HTMLButtonElement{attributes: {…}, innerHTML: "button 1", nodeType: 1, tagName: "button"}
+```
+
+2. Observe context changes and call the `update` function accordingly. For example, we want to change the text to the value of the `clickText` input when it changes.
+
+```typescript
+const input = document.getElementById('clickText');
+input.addEventListener('change', (event) => {
+	instance && instance.update(event.target.value);
```
fbasso, line 82: Is it `focusElementDirective` instead of `instance` ?

It makes me think that it would be better, in my opinion, to have a running example here, even if the code is repeated for explanations (maybe they can also be included in the comments).

The question for the example is : where to put it, as it does not belong to any framework ?

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,120 @@
...
+});
+```
+
+3. Clean up when the `focus-element` gets remove from the DOM. This can be done for example using a `MutationObserver` on the parent DOM element.
+
+```typescript
+const cleanup = (mutations: MutationRecord[]) => {
+	mutations.forEach((mutation) => {
+		for (let entry of mutation.removedNodes.entries()) {
+			if (entry.includes(trackElement) && instance) {
```
fbasso, line 92: instance -> focusElementDirective

docs/01-Headless/03-Directives.md
fbasso, line 82: `instance && instance.update(event.target.value);` -> `instance?.update(event.target.value);`

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,120 @@
...
+				instance.destroy();
+			}
+		}
+	});
+};
+const observer = new MutationObserver(cleanup);
+observer.observe(document.getElementById('content'), {childList: true});
+```
+
+### Angular usage
```
fbasso, line 102: I don't know if it's already possible or not, so it's probably for another PR, but we'll need to have the possibilty to switch sections depending on the selected framework, so that we'll be able to see only the framework part that is specific for the reader.

andreascorti: I see two possibilities here:
1) we implement some custom renderer, but then it will work only on our demo site (like samples).
2) we separate MD either duplicating the common part or creating some utils to merge MDs (but again, it won't display correctly in github) and then show the doc only if the selected framework correspond to the one for which the documentation is written. 

I would rather create an issue only for this, we need to refine the feature.

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,244 @@
 # Directives
 
-## TODO
+## Overview
 
-To be done, issue #401
+Directives in AgnosUI are inspired by the **actions** in Svelte. These directives are element-level lifecycle functions that are executed when the element is created. They are updated (if an update function is provided) when a parameter changes, and finally, they are executed when the element is destroyed.
+These directives are typically used to add custom event handlers. For instance, core services like [focus track](../services/01-Focustrack.md) and [floating ui](../services/03-Floating-UI.md) create directives that are utilized by components such as [Select](../components/select/examples).
```
quentinderoubaix, line 6: If you want the relative links to work in GitHub, you would have to update from `../services/01-Focustrack.md` to `../04-Services/01-Focustrack.md`

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,244 @@
...
+During the `OnChanges` hook, the directive will be executed for the first time (if the reference of the `Directive`) or the Directive `update` function (if provided) will be executed with the new `auUseParams` input.
+
+#### Example
+
+```typescript
+import {UseDirective} from '@agnos-ui/angular-headless';
+import {Component} from '@angular/core';
+import {FormsModule} from '@angular/forms';
+
+@Component({
```
ExFlo, line 118: Maybe we should highlight that we can use the same directive agnostic code for all framework...
Extract the directive code above the three framework ?

docs/01-Headless/03-Directives.md
```diff
@@ -1,5 +1,156 @@
...
+[Directives in React](../../react/docs/Directives.md),
+[Directives in Svelte](../../svelte/docs/Directives.md)
+
+<!-- </framework-specific> -->
+
+## Advanced
+
+### Merge directives
+
+If you are using svelte, you can have multiple directives by using multiple time the `use` directive. On other frameworks, this is not the case.
```
quentinderoubaix, line 149: Very minor: all frameworks should be titlecased, so you could rename here to Svelte (there is also a react one later).

PR #411, opened 2024-01-25
demo: adding API documentation with typedoc in an iframe

divdavem:


quentinderoubaix:
thanks for the initial PR @divdavem !
closing this as this was handled through #852

PR #425, opened 2024-01-29
feat: demo reload after app update

quentinderoubaix:
When we publish a new version of the demo to ```latest```, we do not properly refresh the app.
This is solved by setting a pollInterval value superior to 0 in the svelte.config.js (see [the config doc](https://kit.svelte.dev/docs/configuration#version)).

### DIFF
--- a/demo/src/routes/+layout.svelte
+++ b/demo/src/routes/+layout.svelte
@@ -50,7 +50,7 @@
 </svelte:head>
 -->
 
-<div class="agnos-ui">
+<div class="agnos-ui" data-sveltekit-reload={$updated ? '' : 'off'}>
 	<nav class="demo-nav-top navbar navbar-dark bg-primary bg-gradient">
 		<div class="container-fluid">
 			<a class="navbar-brand" href={$pathToRoot$}> AgnosUI </a>
--- a/demo/svelte.config.js
+++ b/demo/svelte.config.js
@@ -50,6 +50,7 @@ const config = {
 
 		version: {
 			name: child_process.execSync('git rev-parse HEAD').toString().trim(),
+			pollInterval: 5_000,
 		},
 	},
 };


demo/src/routes/+layout.svelte
```diff
@@ -50,7 +50,7 @@
 </svelte:head>
 -->
 
-<div class="agnos-ui">
+<div class="agnos-ui" data-sveltekit-reload={$updated ? '' : 'off'}>
```
fbasso, line 53: I have one question, the only doc that I found about this property is here : https://kit.svelte.dev/docs/link-options#data-sveltekit-reload.

But it applies on a link, and there is no value associated to this attribute. Is there a more precise doc somewhere ?

quentinderoubaix: This is not super documented indeed.
I found this from a Rich Harris presentation + this comment:
https://github.com/sveltejs/kit/discussions/8665#discussioncomment-4760788

While the official doc includes this:
https://kit.svelte.dev/docs/configuration#version

quentinderoubaix: From my understanding, doing ```data-sveltekit-reload={$updated ? '' : 'off'}``` is doing the same thing as
```typescript
beforeNavigate(({willUnload, to}) => {
		if ($updated && !willUnload && to?.url) {
			// force reload of the page on navigation when a new version of the site has been detected
			location.href = to.url.href;
		}
	});
```
but in a more succinct way.
The important point was to include the ```pollInterval``` config to make sure the store ```updated``` is... updated :)

demo/svelte.config.js
```diff
@@ -50,6 +50,7 @@ const config = {
 
 		version: {
 			name: child_process.execSync('git rev-parse HEAD').toString().trim(),
+			pollInterval: 5_000,
```
andreascorti, line 53: Do we really need to poll each 5 sec for an app version? Did you see any performance issue with this option?
It can be useful for devs but I don't think we will release latest so many times to use this option. But if it is negligible the drawback, why not :)

PR #429, opened 2024-01-31
fix(slider): style bootstrap font-weight + margin

ValentinNelu:
Bug in the font weight of the label now, added font-weight for label + used calc for margins


quentinderoubaix (approved):
LGTM !

ValentinNelu:
Thx for the review @quentinderoubaix

PR #434, opened 2024-02-01
feat: pagination hrefs factory function

### LINKED ISSUE #158: Pagination: add the href usage feature
Be able to use the link as real links.

mrednic-1A:
closes #158

### DIFF
--- a/core/src/components/pagination/pagination.spec.ts
+++ b/core/src/components/pagination/pagination.spec.ts
@@ -3,10 +3,12 @@ import {beforeEach, describe, expect, test, vi} from 'vitest';
 import type {PaginationState, PaginationWidget} from './pagination';
 import {createPagination, getPaginationDefaultConfig} from './pagination';
 import {ngBootstrapPagination} from './bootstrap';
+import {assign} from '../../../../common/utils';
 
 describe(`Pagination`, () => {
 	let pagination: PaginationWidget;
 	let state: PaginationState;
+	let expectedState: PaginationState;
 
 	let consoleErrorSpy: MockInstance<Parameters<typeof console.error>, ReturnType<typeof console.error>>;
 
@@ -20,6 +22,7 @@ describe(`Pagination`, () => {
 		const unsubscribe = pagination.state$.subscribe((newState) => {
 			state = newState;
 		});
+		expectedState = state;
 		return () => {
 			unsubscribe();
 			expect(consoleErrorSpy).not.toHaveBeenCalled();
@@ -34,22 +37,36 @@ describe(`Pagination`, () => {
 	};
 
 	test(`should have sensible state`, () => {
-		// TODO we don't test ariaPageLabel here...
-		expect(state).toMatchObject({
+		expect(state).toStrictEqual({
 			pageCount: 1, // total number of page
 			page: 1, // current page
 			pages: [1], // list of the visible pages
 			previousDisabled: true,
+			ariaLabel: 'Page navigation',
+			className: '',
 			nextDisabled: true,
 			disabled: false,
 			directionLinks: true,
 			boundaryLinks: false,
+			slotEllipsis: '…',
+			slotFirst: '«',
+			slotLast: '»',
+			slotNext: '›',
+			slotNumberLabel: state.slotNumberLabel,
+			slotPages: undefined,
+			slotPrevious: '‹',
 			size: null,
 			activeLabel: '(current)',
 			ariaFirstLabel: 'Action link for first page',
 			ariaLastLabel: 'Action link for last page',
 			ariaNextLabel: 'Action link for next page',
 			ariaPreviousLabel: 'Action link for previous page',
+			directionsHrefs: {
+				next: '#',
+				previous: '#',
+			},
+			pagesHrefs: ['#'],
+			pagesLabel: ['Page 1 of 1'],
 		});
 	});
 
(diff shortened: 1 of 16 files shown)


core/src/components/pagination/pagination.ts
```diff
@@ -399,6 +433,22 @@ export function createPagination(config?: PropsConfig<PaginationProps>): Paginat
...
+		const pageCount = pageCount$();
+		return Array.from({length: pageCount}, (_, index) => index + 1).map((page) => pageLinkFactory(page, pageCount));
+	});
+
+	const directionsHrefs$ = computed(() => {
+		const pageLinkFactory = pageLink$();
+		const pageCount = pageCount$();
+		const currentPage = page$();
+		return {
+			previous: pageLinkFactory(currentPage - 1, pageCount),
```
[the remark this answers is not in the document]
mrednic-1A: Indeed I think in can be done taking the corresponding index from the array

core/src/components/pagination/pagination.ts
```diff
@@ -324,6 +353,9 @@ const defaultConfig: PaginationProps = {
 	slotLast: '»',
 	slotPages: undefined,
 	slotNumberLabel: ({displayedPage}) => `${displayedPage}`,
+	pageLink: (_page: number, _pagesCount: number) => {
+		return '#';
+	},
```
fbasso, line 358: Can be simplified with (_page: number, _pagesCount: number) => '#'

angular/demo/src/app/samples/pagination/default.route.ts
```diff
@@ -6,23 +6,46 @@ import {Component} from '@angular/core';
...
-		<nav auPagination auCollectionSize="60" [(auPage)]="pageAlone" auAriaLabel="Disabled pagination" auDisabled></nav>
+		<nav auPagination auCollectionSize="60" [(auPage)]="pageAlone" [auPageLink]="pageLink" auAriaLabel="Disabled pagination" auDisabled></nav>
 	`,
 })
 export default class DefaultPaginationComponent {
 	page = 4;
 	pageAlone = 1;
+
+	constructor() {
+		let jsonObj: {page?: number} = {};
```
ExFlo, line 30: Is it possible to export this logic in an angular service ?

core/src/components/pagination/pagination.ts
```diff
@@ -448,6 +448,15 @@ export function createPagination(config?: PropsConfig<PaginationProps>): Paginat
 		};
 	});
 
+	function handleNavigation(pageNumber: number, event?: Event & {ctrlKey: boolean; metaKey: boolean}) {
+		if (!event || (!event.ctrlKey && !event.metaKey)) {
+			setTimeout(() => patch({page: pageNumber}));
```
ExFlo, line 453: do we need the setTimeout here ?

angular/demo/src/app/samples/pagination/default.route.ts
mrednic-1A: I could do it in an angular service, or I could adapt [this function](https://github.com/AmadeusITGroup/AgnosUI/blob/2a92391c534dde7356367de324b8f246e1811f51/common/utils.ts#L17)
 to return the full JSON object, and use it in the other apps in the example.
 
BTW I was thinking to do another PR for the example ( and remove it from this one just before merge).
what do you think ?

core/src/components/pagination/pagination.ts
mrednic-1A: it is needed only for 'next' and 'previous' just because the patch it is changing the urls, before the browser is changing the url when pressing the `a`.
as example
initial state: `currentPage: 4`, `nextHref: #{page:5}`, browser url: url#{page:4}
click on 'next' -> `currentPage: 5`, `nextHref: #{page:6}`, browser url: url#{page:6} -instead of 5
skipping one tick it's updating the browser href, before the current page which updates the next

angular/demo/src/app/samples/pagination/default.route.ts
fbasso: As it's a simple example, maybe you can simplify the way it's managed with the hash. For example, you can just manage an fragment like this :

`#${hashFragment}#${_pageNumber}}`

It would simplify the logic of the example.

By the way, it's no easy for the user to know what's happened here, because we are changing (of course) the url of the iframe, but it's not directly visible. I don't know how it have it clearer... (specific example where the hash is output on the display ?)

ExFlo: I don't know about the function. maybe we need to discuss if this is interesting to share it. If not service. Bring this into a meeting.

core/src/components/pagination/pagination.ts
fbasso: Good catch for this issue ! 

I don't know if it's a good solution or not, and what kind of checks that need to be done, but the only way I found to keep it synchronous is finally to manage the href internally, like this:

```typescript
		if (!event || (!event.ctrlKey && !event.metaKey)) {
			patch({page: pageNumber});
			const href = (event?.target as HTMLAnchorElement)?.href;
			if (href) {
				history.pushState(null, '', href);
				event?.preventDefault();
			}
		}

```

mrednic-1A: Thanks :) I'll give it a try .

angular/demo/src/app/samples/pagination/default.route.ts
mrednic-1A: After discussions the examples have been a lot simplified. There is a small common part which exposes the hash as a store.

core/src/components/pagination/pagination.spec.ts
```diff
@@ -17,6 +18,7 @@ describe(`Pagination`, () => {
 		// Need to move those test to only ngBootstrapPagination tests
 		// Need to add test of the simple widget alone.
 		pagination.patch({pagesFactory: ngBootstrapPagination(0, false, true)});
+		expectedState = pagination.state$();
```
fbasso, line 21: If you want to manage the expectedState globally (as you do. Another solution would have been to crerate this variable in each test), it's better to do `expectedState = state` just after the following subscribe. (`pagination.state$()` do a subscribe/unsubscribe to get the value, this would avoid it)

core/src/components/pagination/pagination.ts
```diff
@@ -399,6 +435,36 @@ export function createPagination(config?: PropsConfig<PaginationProps>): Paginat
...
+	 * Update page number when navigation is in the same tab and stop the event propagation;
+	 * For navigations outside current browser tab let the default behavior, without updating the page number;
+	 */
+	function handleNavigation(pageNumber: number, event?: PaginationClickEvent) {
+		if (pagesHrefs$()[pageNumber - 1] === PAGE_LINK_DEFAULT) {
+			event?.preventDefault();
+		}
+		if (!event || !(event.ctrlKey || event.metaKey)) {
+			event?.preventDefault();
+			patch({page: pageNumber});
```
fbasso, line 464: You can rename `pageNumber` to `page` and just do `patch({page});`. (I know, it's micro optimization, but it's for free ;) )

e2e/pagination/pagination.e2e-spec.ts
```diff
@@ -13,14 +14,17 @@ interface returnState {
 
 async function paginationState(paginationPO: PaginationPO) {
 	return paginationPO.locatorRoot.evaluate((rootNode: HTMLElement) => {
-		const returnState: returnState = {rootClasses: [], disabled: null, pages: []};
+		const returnState: returnState = {rootClasses: [], disabled: null, pages: [], hrefs: []};
 		const pagesElements = [...rootNode.querySelectorAll('.au-page')] as HTMLLinkElement[];
 		const pages = [];
+		const hrefs = [];
 		for (const element of pagesElements) {
+			hrefs.push(element.getAttribute('href') || ''.trim());
```
fbasso, line 22: Missing parenthesis, otherwise not sure that the `trim` is applied to your attribute. But is it necessary to trim here ?

e2e/pagination/pagination.e2e-spec.ts
```diff
@@ -5,6 +5,7 @@ interface returnState {
 	rootClasses: string[];
 	disabled: string | null;
 	pages: string[];
+	hrefs: string[];
```
fbasso, line 8: The hrefs for the boundaries are separated, so I would add some properties for them, to be included in the different tests

e2e/pagination/pagination.e2e-spec.ts
```diff
@@ -121,4 +150,28 @@ test.describe.parallel(`Pagination tests`, () => {
...
+				'#/pagination/hash#3',
+				'#/pagination/hash#4',
+				'#/pagination/hash#5',
+				'#/pagination/hash#6',
+			],
+		};
+		const paginationPO = new PaginationPO(page, 0);
+		await page.goto('#/pagination/hash');
+		await paginationPO.locatorRoot.waitFor();
+		await expect.poll(() => paginationState(paginationPO)).toEqual(stateWithCustomHrefs);
```
fbasso, line 175: Maybe we can test more, by including the click on one page and check the state (it will make sense if the boundaries href are included), and click on the browser history go back (I think there is a specific instruction with Playwright)

core/src/components/pagination/pagination.ts
```diff
@@ -232,8 +232,33 @@ export interface PaginationProps extends PaginationCommonPropsAndState {
...
 	 */
 	ariaPageLabel: (processPage: number, pageCount: number) => string;
+
+	/**
+	 * Factory function providing the href for a "Page" page anchor,
+	 * based on the current page number
+	 * @param pageNumber - The index to use in the link
+	 * @defaultValue
+	 * ```ts
+	 * ({pageNumber}) => `?page=${pageNumber}`
```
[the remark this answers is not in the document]
mrednic-1A: indeed

core/src/components/pagination/pagination.ts
```diff
@@ -232,8 +232,33 @@ export interface PaginationProps extends PaginationCommonPropsAndState {
...
+	 */
+	previous: string;
+	/**
+	 * The href for the 'Next' direction link
+	 */
+	next: string;
+}
+
+/** Type of the event fired when a pagination slot is clicked */
+export type PaginationClickEvent = Pick<PointerEvent, 'ctrlKey' | 'metaKey' | 'preventDefault'>;
```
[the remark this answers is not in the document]
mrednic-1A: As it is in the core, when calling the function from react component, I have a miss match on types. We can have a look together

e2e/pagination/pagination.e2e-spec.ts
mrednic-1A: I think you are right in both cases. :)
I will remove it.

e2e/pagination/pagination.e2e-spec.ts
mrednic-1A: indeed

e2e/pagination/pagination.e2e-spec.ts
mrednic-1A: indeed. nice point for the browser history.

core/src/components/pagination/pagination.ts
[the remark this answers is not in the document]
fbasso: It is `(_page: number) => '#'` ;)

fbasso:
LGTM, thanks @mrednic-1A !

PR #450, opened 2024-02-01
fix: mobile tabs in header

fbasso:
another way to present tabs, to be discussed...


fbasso:
Close as we need to find a better way of managing how tabs are displayed, in mobile.

PR #462, opened 2024-02-05
test: add tests and fix accessibility issues for the select

### LINKED ISSUE #263: Accessibility documentation for select component
**Is your feature request related to a problem? Please describe.**
Like all components, the documentation of the select should contain how we make sure the widget is accessible and what are the aria fields that can be set.

### LINKED ISSUE #459: select: add accessibility tests

fbasso:
fix #459
fix #263

### DIFF
--- a/angular/lib/src/components/select/select.component.ts
+++ b/angular/lib/src/components/select/select.component.ts
@@ -44,6 +44,7 @@ export class SelectItemDirective<Item> {
 				[auUse]="controlContainerDirective"
 				role="combobox"
 				class="d-flex align-items-center flex-wrap"
+				[attr.aria-controls]="state.id + '-menu'"
 				aria-haspopup="listbox"
 				[attr.aria-expanded]="state.open"
 			>
@@ -75,6 +76,8 @@ export class SelectItemDirective<Item> {
 			</div>
 			@if (state.open && state.visibleItems.length) {
 				<ul
+					role="listbox"
+					[id]="state.id + '-menu'"
 					[auUse]="menuDirective"
 					[class]="'dropdown-menu show ' + (menuClassName || '')"
 					[attr.data-popper-placement]="state.placement"
@@ -82,7 +85,10 @@ export class SelectItemDirective<Item> {
 				>
 					@for (itemContext of state.visibleItems; track itemCtxTrackBy($index, itemContext)) {
 						<li
-							[class]="'au-select-item dropdown-item position-relative' + (itemContext === state.highlighted ? ' bg-primary text-light' : '')"
+							role="option"
+							[attr.aria-selected]="itemContext.selected"
+							class="au-select-item dropdown-item position-relative"
+							[class.text-bg-primary]="itemContext === state.highlighted"
 							[class.selected]="itemContext.selected"
 							(click)="widget.api.toggleItem(itemContext.item)"
 						>
--- a/core/src/utils/internal/dom.ts
+++ b/core/src/utils/internal/dom.ts
@@ -87,3 +87,11 @@ export function addEvent(element: Element, type: string, fn: EventListenerOrEven
 		element.removeEventListener(type, fn);
 	};
 }
+
+let idCount = 0;
+/**
+ * Generates a unique ID with the format 'auId-[counter]'.
+ *
+ * @returns The generated ID.
+ */
+export const generateId = () => `auId-${idCount++}`;
--- a/e2e/select/select.chromium.e2e-spec.ts
+++ b/e2e/select/select.chromium.e2e-spec.ts
@@ -1,0 +1,22 @@
+import {SelectPO} from '@agnos-ui/page-objects';
+import {expect, test} from '../fixture';
+import {analyze, switchTheme} from '../utils';
+
+test.describe('Accessibility', () => {
+	test('Opened select', async ({page, browserName}) => {
+		if (browserName === 'chromium') {
+			await page.goto('#/select/default');
+			const selectPO = new SelectPO(page);
+			const locatorInput = selectPO.locatorInput;
+			await locatorInput.fill('a');
+			await locatorInput.press('Enter');
+			await expect.poll(async () => (await selectPO.state()).badges.length).toBeGreaterThan(0);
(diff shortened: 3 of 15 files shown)


angular/lib/src/components/select/select.component.ts
```diff
@@ -75,14 +76,16 @@ export class SelectItemDirective<Item> {
...
 					[auUse]="menuDirective"
 					[class]="'dropdown-menu show ' + (menuClassName || '')"
 					[attr.data-popper-placement]="state.placement"
 					(mousedown)="$event.preventDefault()"
 				>
 					@for (itemContext of state.visibleItems; track itemCtxTrackBy($index, itemContext)) {
 						<li
-							[class]="'au-select-item dropdown-item position-relative' + (itemContext === state.highlighted ? ' bg-primary text-light' : '')"
+							[class]="'au-select-item dropdown-item position-relative'"
+							[class.bg-text-bg-primary]="itemContext === state.highlighted"
```
ExFlo, line 88: is this this good in dark mode ?

fbasso: I don't know what's happened, but the point is that the bg-text-bg-primary class is not included anymore :(

And it should be fine for the dark mode, as it's like this Bootstrap manages it (there is no color switch in bootstrap. For example https://getbootstrap.com/docs/5.3/components/toasts/#color-schemes).

@ExFlo, how is it supposed to be done ?

fbasso: Oh, my bad, it's `text-bg-primary` that needs to be used !

angular/lib/src/components/select/select.component.ts
```diff
@@ -75,14 +76,19 @@ export class SelectItemDirective<Item> {
...
 					[class]="'dropdown-menu show ' + (menuClassName || '')"
 					[attr.data-popper-placement]="state.placement"
 					(mousedown)="$event.preventDefault()"
 				>
 					@for (itemContext of state.visibleItems; track itemCtxTrackBy($index, itemContext)) {
 						<li
-							[class]="'au-select-item dropdown-item position-relative' + (itemContext === state.highlighted ? ' bg-primary text-light' : '')"
+							role="option"
+							[attr.aria-selected]="itemContext.selected"
+							[class]="'au-select-item dropdown-item position-relative'"
```
quentinderoubaix, line 90: Could be just
```class="au-select-item dropdown-item position-relative"```

File: core/src/utils/internal/dom.ts [hunk 3]
```diff
@@ -87,3 +87,11 @@ export function addEvent(element: Element, type: string, fn: EventListenerOrEven
...
 	};
 }
+
+let idCount = 0;
+/**
+ * Generates a unique ID with the format 'auId-[counter]'.
+ *
+ * @returns The generated ID.
+ */
+export const getGeneratedId = () => `auId-${idCount++}`;
```
quentinderoubaix, line 97: maybe rename to **generateId** ?
I usually prefer to add the *get* prefix only when the rest of the name does not clarify that it returns something.

another thing: after merging this, create an issue to use this in the accordion item, which also manages a unique id.

e2e/select/select.chromium.e2e-spec.ts
```diff
@@ -0,0 +1,16 @@
...
+import {analyze} from '../utils';
+
+test.describe('Accessibility', () => {
+	test('Opened select', async ({page, browserName}) => {
+		if (browserName === 'chromium') {
+			await page.goto('#/select/default');
+			const selectPO = new SelectPO(page);
+			const locatorInput = selectPO.locatorInput;
+			await locatorInput.fill('a');
+			await locatorInput.press('Enter');
```
quentinderoubaix, line 12: maybe add an expect before the axe tests to make sure the popup is opened.

e2e/select/select.chromium.e2e-spec.ts
```diff
@@ -0,0 +1,16 @@
...
+			expect((await analyze(page, page.url())).violations).toEqual([]);
```
quentinderoubaix, line 13: Note that in the main accessibility test, we test for both light and dark modes.
I wonder if we should do the same here...

svelte/lib/src/components/select/Select.svelte
```diff
@@ -55,15 +55,17 @@
 		directives: {floatingDirective, hasFocusDirective, referenceDirective, inputContainerDirective},
 	} = widget;
 	$: widget.patchChangedProps($$props);
+
+	$: menuId = `${$id$}-menu`;
```
quentinderoubaix, line 59: a lot of dollars 😄

quentinderoubaix (commented):
@fbasso left some minor comments, thx for the PR !

angular/lib/src/components/select/select.component.ts
fbasso: Right :)

e2e/select/select.chromium.e2e-spec.ts
fbasso: Done ;). I've added more by the way:

- the first one to be sure the first item is selected,
- the second one to be sure that the dropdown is still open.

fbasso:
@quentinderoubaix, pushed  again with changes.

e2e/utils.ts
```diff
@@ -1,4 +1,30 @@
...
+ * @param page
+ * @param route
+ * @returns
+ */
+export async function analyze(page: Page, route: string): Promise<AxeResults> {
+	const analyser = new AxeBuilder({page}).withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa']);
+	return analyser.analyze();
+}
+
+export async function swithTheme(page: Page, theme: 'light' | 'dark') {
```
quentinderoubaix, line 21: little typo, should be ```switchTheme```

quentinderoubaix (approved):
thanks for the changes @fbasso !

e2e/utils.ts
fbasso: Thanks !

fbasso:
Thanks @quentinderoubaix

PR #470, opened 2024-02-07
feat: refactor directives purposes

### LINKED ISSUE #537: core: refactor directives purpose
This will:

- Give more responsibilities to the directives, to add events, attributes, ...
- Be able to get the right markup server side

fbasso:
fix #537

This is a POC to see what's happen if we delegate more works on directives:

- increase the common code
- decrease the code duplication between frameworks
- ease the headless usage

On the other hand:

-  bypass the framework templating  engine

### DIFF
--- a/core/src/components/rating/rating.ts
+++ b/core/src/components/rating/rating.ts
@@ -1,11 +1,13 @@
+import type {ReadableSignal} from '@amadeus-it-group/tansu';
 import {computed, writable} from '@amadeus-it-group/tansu';
+import type {ConfigValidator, Directive, PropsConfig, SlotContent, Widget} from '../../types';
 import {INVALID_VALUE} from '../../types';
-import {bindableProp, stateStores, writablesForProps} from '../../utils/stores';
+import {createAttributesDirective} from '../../utils/directive';
 import {clamp, isNumber} from '../../utils/internal/checks';
-import {typeBoolean, typeFunction, typeNumber, typeString} from '../../utils/writables';
-import type {ConfigValidator, PropsConfig, SlotContent, Widget} from '../../types';
-import type {WidgetsCommonPropsAndState} from '../commonProps';
 import {noop} from '../../utils/internal/func';
+import {bindableProp, stateStores, writablesForProps} from '../../utils/stores';
+import {typeBoolean, typeFunction, typeNumber, typeString} from '../../utils/writables';
+import type {WidgetsCommonPropsAndState} from '../commonProps';
 
 export interface StarContext {
 	/**
@@ -98,6 +100,19 @@ export interface RatingProps extends RatingCommonPropsAndState {
 	onLeave: (rating: number) => void;
 }
 
+export interface RatingDirectives {
+	/**
+	 * A directive to be applied to the main container
+	 * This will handle the keydown, mouseleave, tabindex and aria attributes
+	 */
+	containerDirective: Directive;
+
+	/**
+	 * A directive to be applied on each star element
+	 */
+	starDirective: Directive<{index: number}>;
+}
+
 export interface RatingState extends RatingCommonPropsAndState {
 	/**
 	 * the aria value of the rating
@@ -149,7 +164,7 @@ export interface RatingActions {
 	handleKey(event: KeyboardEvent): void;
 }
 
-export type RatingWidget = Widget<RatingProps, RatingState, object, RatingActions>;
+export type RatingWidget = Widget<RatingProps, RatingState, object, RatingActions, RatingDirectives>;
 
 const defaultConfig: RatingProps = {
 	rating: 0,
@@ -205,9 +220,7 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
 			rating$: _dirtyRating$,
 			tabindex$: _dirtyTabindex$,
 
-			// clean inputs with value validation:
 			ariaValueTextFn$,
-
 			onHover$,
 			onLeave$,
 			onRatingChange$,
@@ -216,7 +229,8 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
 		},
 		patch,
 	] = writablesForProps(defaultConfig, config, configValidator);
-	const {maxRating$, disabled$, readonly$, resettable$} = stateProps;
+
+	const {ariaLabel$, ariaLabelledBy$, className$, disabled$, maxRating$, readonly$, resettable$} = stateProps;
 
 	// clean inputs adjustment to valid range
 	const tabindex$ = computed(() => (disabled$() ? -1 : _dirtyTabindex$()));
@@ -241,7 +255,7 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
 		}));
 	});
 
-	return {
+	const widget: RatingWidget = {
 		...stateStores({
 			ariaValueText$,
 			interactive$,
@@ -298,7 +312,46 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
 				}
 			},
 		},
-		directives: {},
+		directives: {
+			containerDirective: createAttributesDirective(() => ({
+				events: {
+					keydown: (e) => widget.actions.handleKey(e),
+					mouseleave: () => widget.actions.leave(),
+				},
+				attributes: {
+					role: 'slider',
+					class: className$,
+					'aria-valuemin': 0,
+					tabindex: tabindex$,
+					'aria-valuemax': maxRating$,
+					'aria-valuenow': visibleRating$,
+					'aria-valuetext': ariaValueText$,
+					'aria-readonly': computed(() => (readonly$() ? 'true' : undefined)),
+					'aria-disabled': computed(() => (disabled$() ? 'true' : undefined)),
+					'aria-label': ariaLabel$,
+					'aria-labelledby': computed(() => ariaLabelledBy$() || undefined),
+				},
+				classNames: {
+					'au-rating': true,
+				},
+			})),
+			starDirective: createAttributesDirective((starContext$: ReadableSignal<{index: number}>) => {
+				return {
+					events: {
+						mouseenter: () => widget.actions.hover(starContext$().index + 1),
+						click: () => widget.actions.click(starContext$().index + 1),
+					},
+					styles: {
+						cursor: computed(() => (interactive$() ? 'pointer' : 'default')),
+					},
+					classNames: {
+						'au-rating-star': true,
+					},
+				};
+			}),
+		},
 		api: {},
 	};
+
+	return widget;
 }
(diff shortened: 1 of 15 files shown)


fbasso:
Ready for review, but I wonder that we should maybe remove the 'action' from the public api, as there are supposed to be used in the directives directly, so there is no need to be outside (and in the future, being customizable through the directive parameters).

core/src/utils/directive.ts
```diff
@@ -223,3 +227,124 @@ export const mergeDirectives =
...
+
+/**
+ * Creates a directive for server-side rendering with bindable elements.
+ * This directive binds events, attributes, styles, and classNames to an HTML element.
+ *
+ * @param propsFn - A function that returns the SSRDirectiveProps with the data to bind.
+ * This function can take an optional parameter that corrspond to the second parameter of the created directive.
+ * @returns A directive object with bound events, attributes, styles, and classNames.
+ */
+export const createSSRDirective =
```
divdavem, line 272:
```suggestion
export const createAttributesDirective =
```
This is not really linked to SSR.

core/src/utils/directive.ts
```diff
@@ -223,3 +227,124 @@ export const mergeDirectives =
...
+	<T = void>(propsFn: (arg: T) => SSRDirectiveProps) =>
+	(node: HTMLElement, args: T) => {
+		const {events, attributes, styles, classNames} = propsFn(args);
```
divdavem, line 275: As discussed, I think `args` should be put in a store and passed to the `propsFn` function as a store, and the directive should provide an update function that changes that store..

core/src/utils/internal/dom.ts
```diff
@@ -95,3 +99,136 @@ let idCount = 0;
...
+				classNames.delete(className);
+			}
+		},
+	};
+
+	const node = {
+		addEventListener: noop,
+		removeEventListener: noop,
+		setAttribute(name: string, value: string) {
+			attributes[name] = value;
```
divdavem, line 133: I think setting the `class` and `style` attributes should work in the virtual node just like in the browser.

svelte/lib/src/components/rating/Rating.svelte
```diff
@@ -26,52 +27,21 @@
...
-	aria-valuenow={$visibleRating$}
-	aria-valuetext={$ariaValueText$}
-	aria-readonly={$readonly$ || undefined}
-	aria-disabled={$disabled$ || undefined}
-	aria-label={$ariaLabel$ || undefined}
-	aria-labelledby={$ariaLabelledBy$ || undefined}
-	on:keydown={handleKey}
-	on:mouseleave={leave}
->
+<div use:containerDirective {...containerAttributes} class="d-inline-flex au-rating {$className$}">
```
divdavem, line 39: How would you integrate the `class` attribute in containerDirective?

core/src/utils/directive.ts
```diff
@@ -1,7 +1,11 @@
 import type {ReadableSignal} from '@amadeus-it-group/tansu';
 import {asReadable, batch, readable, writable} from '@amadeus-it-group/tansu';
+import {BROWSER} from 'esm-env';
```
divdavem, line 3: If you use `esm-env`, you should add it to the peer dependencies (or dependencies) of the core and check it is part of the external rollup options in `vite.config.ts`. Otherwise, vite will bundle it with the core with `BROWSER` hard-coded to `true` and it no longer does what you expect.

svelte/lib/src/components/rating/Rating.svelte
```diff
@@ -26,52 +27,21 @@
...
-	aria-valuenow={$visibleRating$}
-	aria-valuetext={$ariaValueText$}
-	aria-readonly={$readonly$ || undefined}
-	aria-disabled={$disabled$ || undefined}
-	aria-label={$ariaLabel$ || undefined}
-	aria-labelledby={$ariaLabelledBy$ || undefined}
-	on:keydown={handleKey}
-	on:mouseleave={leave}
->
+<div use:containerDirective {...containerAttributes} class="d-inline-flex">
```
divdavem, line 39: Having both `{...containerAttributes}` (which includes `class` in SSR) and `class="d-inline-flex` will not work well, I think.

core/src/components/rating/rating.ts
```diff
@@ -298,7 +320,44 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
...
+			starDirective: createAttributesDirective((starContext$: ReadableSignal<{index: number}>) => {
+				const index = starContext$().index;
+				return {
+					events: {
+						mouseenter: () => widget.actions.hover(index + 1),
+						click: () => widget.actions.click(index + 1),
+					},
+					styles: {
+						cursor: computed(() => (interactive$() ? 'pointer' : 'default')),
+					},
```
divdavem, line 355: Should we have the `au-rating-star` class here, as we have `au-rating` in `containerDirective`?

react/lib/src/components/rating/rating.tsx
```diff
@@ -1,49 +1,37 @@
...
-					<span
-						className="au-rating-star"
-						style={starStyle}
-						onMouseEnter={() => widget.actions.hover(star.index + 1)}
-						onClick={() => widget.actions.click(star.index + 1)}
-					>
-						<Slot slotContent={slotStar} props={star}></Slot>
-					</span>
-				</React.Fragment>
+		<div ref={refContainer} className="d-inline-flex">
```
divdavem, line 32: Shouldn't we use `ssrDirectiveAttributes` also in react? Or should we wait for a preprocessor to add it everywhere automatically?

divdavem: Note that we would need a specific utility function for react to produce react-specific attributes like `className` from the result of `ssrDirectiveAttributes`. This should probably be done later in another PR.

core/src/components/rating/rating.ts
```diff
@@ -298,7 +320,47 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
...
+					'aria-valuemax': maxRating$,
+					'aria-valuenow': visibleRating$,
+					'aria-valuetext': ariaValueText$,
+					'aria-readonly': computed(() => (readonly$() ? 'true' : undefined)),
+					'aria-disabled': computed(() => (disabled$() ? 'true' : undefined)),
+					'aria-label': ariaLabel$,
+					'aria-labelledby': computed(() => ariaLabelledBy$() || undefined),
+				},
+				classNames: {
+					'au-rating': writable(true),
```
divdavem, line 343:
```suggestion
					'au-rating': true,
```

core/src/components/rating/rating.ts
```diff
@@ -298,7 +320,47 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
...
+				return {
+					events: {
+						mouseenter: () => widget.actions.hover(index + 1),
+						click: () => widget.actions.click(index + 1),
+					},
+					styles: {
+						cursor: computed(() => (interactive$() ? 'pointer' : 'default')),
+					},
+					classNames: {
+						'au-rating-star': writable(true),
```
divdavem, line 357:
```suggestion
						'au-rating-star': true,
```

core/src/components/rating/rating.ts
```diff
@@ -298,7 +320,47 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
...
+					'aria-disabled': computed(() => (disabled$() ? 'true' : undefined)),
+					'aria-label': ariaLabel$,
+					'aria-labelledby': computed(() => ariaLabelledBy$() || undefined),
+				},
+				classNames: {
+					'au-rating': writable(true),
+				},
+			})),
+			starDirective: createAttributesDirective((starContext$: ReadableSignal<{index: number}>) => {
+				const index = starContext$().index;
```
divdavem, line 347: `starContext$` is a readable signal, its value could change over time. If you evaluate it here, and the value changes later, the new value will not be used. You should evaluate `starContext$()` only when using it.

react/lib/src/components/rating/rating.tsx
```diff
@@ -1,49 +1,37 @@
...
 import React from 'react';
 
 export * from '@agnos-ui/react-headless/components/rating';
 
+function Star({star, state, directive}: {star: StarContext; state: RatingState; directive: RatingDirectives['starDirective']}) {
+	const starContainer = useDirective(directive, {index: star.index});
+	return (
+		<React.Fragment key={star.index}>
+			<span className="visually-hidden">({star.index < state.visibleRating ? '*' : ' '})</span>
+			<span ref={starContainer} className="au-rating-star">
```
divdavem, line 15:
```suggestion
			<span ref={starContainer}>
```
`au-rating-star` is already added by the `starContainer` directive, right?

core/src/components/rating/rating.ts
```diff
@@ -208,9 +223,13 @@ export function createRating(config?: PropsConfig<RatingProps>): RatingWidget {
 			// clean inputs with value validation:
 			ariaValueTextFn$,
 
+			ariaLabel$,
+			ariaLabelledBy$,
+
```
divdavem, line 228: Instead of adding those stores both here and when calling `stateStores()`, you should instead do as for the `maxRating$, disabled$, readonly$, resettable$` stores:
```ts
const {maxRating$, disabled$, readonly$, resettable$, ariaLabel$, ariaLabelledBy$, className$} = stateProps; 
```

core/src/utils/directive.ts
```diff
@@ -223,3 +225,80 @@ export const mergeDirectives =
...
+/**
+ * Properties for configuring server-side rendering directives.
+ */
+export interface AttributesDirectiveProps {
+	/**
+	 * Events to be attached to an HTML element.
+	 * @remarks
+	 * Key-value pairs where keys are event types and values are event handlers.
+	 */
+	events?: Partial<Record<keyof HTMLElementEventMap, (event: any) => void>>;
```
divdavem, line 238:
```suggestion
	events?: Partial<{[K in keyof HTMLElementEventMap]: (this: HTMLElement, event: HTMLElementEventMap[K]) => void}>;
```

core/src/utils/directive.ts
```diff
@@ -223,3 +225,80 @@ export const mergeDirectives =
...
+
+	/**
+	 * Class names to be added to an HTML element.
+	 * @remarks
+	 * Key-value pairs where keys are class names and values indicate whether the class should be added (true) or removed (false).
+	 */
+	classNames?: Record<string, boolean | ReadableSignal<boolean>>;
+}
+
+const getStore = <T>(value: T | ReadableSignal<T>) => (isStore(value) ? value : writable(value));
```
divdavem, line 262: Instead of using this function, you should use `toReadableStore` from `stores.ts`.
I think you only need a readable, not a writable. If you need a writable, `toWritableStore` exists as well in `stores.ts`.

divdavem (approved):
Thank you for your PR

fbasso:
Thanks @divdavem  !

PR #473, opened 2024-02-09
test: add callback events cases

fbasso:
PR to discuss about controlled/uncontrolled widget and when the events must be triggers.

I just added two test for now. It simulates a select, where the user inputs are changed in uppercase, and monitor the `onTextFilterChange` and the `onOpenChange`  callbacks.

Run the test with `npm run tdd:ui core select` and have a look at the expect statements.


quentinderoubaix:
Opened #1062 for future discussions on the subject.

PR #498, opened 2024-02-16
fix: theme icon on first render

quentinderoubaix:


demo/src/routes/menu/Theme.svelte
```diff
@@ -79,7 +80,9 @@
 			data-bs-toggle="dropdown"
 			aria-expanded={$open$}
 		>
-			<Svg svg={$currentTheme$.icon} />
```
andreascorti, line 82: Maybe you can remove `icon` from the `currentTheme$` then, as it is not used anymore

andreascorti (approved):
LGTM @quentinderoubaix, left a very small comment, but you can merge without it

quentinderoubaix:
Thanks for the review @andreascorti !

demo/src/routes/menu/Theme.svelte
quentinderoubaix: Indeed, maybe the currentTheme$ should just be a ```writable(string)```,  and would be the id of the current theme.

PR #507, opened 2024-02-20
feat: add slots to slider

### LINKED ISSUE #464: Add slots to the Slider widget
**Description**
Enhance the Slider component to support slots.

andreascorti:
closes #464 

- [x] Add a full custom structure example

### DIFF
--- a/angular/demo/src/app/samples/slider/custom.route.ts
+++ b/angular/demo/src/app/samples/slider/custom.route.ts
@@ -1,0 +1,34 @@
+import {SliderComponent, SliderLabelDirective} from '@agnos-ui/angular';
+import {Component} from '@angular/core';
+import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
+import CustomHandleSliderComponent from './custom-handle.component';
+
+@Component({
+	standalone: true,
+	imports: [SliderComponent, SliderLabelDirective, ReactiveFormsModule, FormsModule],
+	template: `
+		<h2>Slider with custom labels</h2>
+		<div auSlider auMin="0" auMax="100" auStepSize="1" [formControl]="sliderCustomLabelControl" auClassName="au-custom-slider-container">
+			<ng-template auSliderLabel let-value="value">{{ value }}%</ng-template>
+		</div>
+		<hr />
+
+		<h2>Slider with custom handle</h2>
+		<div
+			auSlider
+			auMin="0"
+			auMax="100"
+			auStepSize="1"
+			[formControl]="sliderCustomHandleControl"
+			auClassName="au-custom-slider-container"
+			[auSlotHandle]="sliderCustomSlotHandle"
+		></div>
+	`,
+	styles: "@import '@agnos-ui/common/samples/slider/custom.scss';",
+})
+export default class CustomSliderComponent {
+	readonly sliderCustomSlotHandle = CustomHandleSliderComponent;
+
+	sliderCustomLabelControl = new FormControl([70]);
+	sliderCustomHandleControl = new FormControl([50]);
+}
--- a/angular/lib/src/components/slider/slider.component.ts
+++ b/angular/lib/src/components/slider/slider.component.ts
@@ -1,6 +1,16 @@
-import type {SliderWidget} from '@agnos-ui/angular-headless';
+import type {
+	SliderContext,
+	SliderProps,
+	SliderSlotHandleContext,
+	SliderSlotLabelContext,
+	SliderWidget,
+	SlotContent,
+} from '@agnos-ui/angular-headless';
 import {
 	BaseWidgetDirective,
+	ComponentTemplate,
+	SlotDefaultDirective,
+	SlotDirective,
 	UseDirective,
 	auBooleanAttribute,
 	auNumberAttribute,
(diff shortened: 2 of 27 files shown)


core/src/components/slider/slider.spec.ts
```diff
@@ -68,6 +71,7 @@ describe(`Slider basic`, () => {
 		const maxLabelDirective = slider.directives.maxLabelDirective(maxLabel);
 		const unsubscribe = slider.state$.subscribe((newState) => {
 			state = newState;
+			defaultStateValues.slotLabel = newState.slotLabel;
```
andreascorti, line 74: @fbasso : we need to discuss how we want to test slots. In this use case, I did the same as pagination, assuring the default function provided by the widget is the same as the expected by our tests. 

But I also noticed in select, you `normalize` the state replacing functions by a string `(function)`, hence you don't check them.

andreascorti: As discussed, I changed it to use a computed state with some manual modification (in this case, changing the function for a string `'(function')`, and changing the default config to a function to avoid cloning the object all the time.

angular/lib/src/components/slider/slider.component.ts
```diff
@@ -1,115 +1,210 @@
...
-	providers: [{provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => SliderComponent), multi: true}],
-	imports: [UseDirective],
-	host: {
-		class: `au-slider`,
-		'[class]': '(state().vertical ? "au-slider-vertical" : "au-slider-horizontal") + " " + state().className',
-		'[class.disabled]': 'state().disabled',
-		'[attr.aria-disabled]': 'state().disabled ? true : null',
-		'(blur)': 'handleBlur()',
-	},
+	imports: [NgClass, SlotDirective, SliderHandleDirective, UseDirective],
```
MarkoOleksiyenko, line 57: don't think that you need the ngClass here, do you?

angular/lib/src/components/slider/slider.component.ts
```diff
@@ -1,115 +1,210 @@
...
+	}
+}
+
+@Directive({selector: 'ng-template[auSliderHandle]', standalone: true})
+export class SliderHandleDirective {
+	public templateRef = inject(TemplateRef<SliderContext>);
+	static ngTemplateContextGuard(
+		_dir: SliderHandleDirective,
+		context: unknown,
+	): context is SliderContext & {item: {value: number; id: number; ariaLabel: string}} {
```
MarkoOleksiyenko, line 49: minor (if you don't change it I will add it in aria-valuetext PR): could be replaced by new interface (SliderHandle) in core

core/src/components/slider/slider.ts
```diff
@@ -97,6 +99,21 @@ export interface SliderCommonPropsAndState extends WidgetsCommonPropsAndState {
 	 * It `true` slider display is inversed
 	 */
 	rtl: boolean;
+
+	/**
+	 * Slot to change the default element of the slider
```
MarkoOleksiyenko, line 104: mb change "default element" to "default display"?

demo/src/routes/docs/[framework]/components/slider/examples/+page.svelte
```diff
@@ -19,6 +21,12 @@
 <Section label="Right to left slider" level={2}>
 	<Sample title="Right to left slider" sample={rightToLeftSample} height={627} />
 </Section>
+<Section label="Custom slider" level={2}>
+	<Sample title="Custom slider" sample={customSample} height={627} />
```
Comment (MarkoOleksiyenko) [inline demo/src/routes/docs/[framework]/components/slider/examples/+page.svelte:25]: minor: you can probably reduce the height of the iframe here

svelte/lib/src/components/slider/SliderDefaultStructure.svelte
```diff
@@ -0,0 +1,130 @@
+<script lang="ts">
+	import type {SliderSlots} from '@agnos-ui/svelte-headless/components/slider';
+	import {Slot} from '@agnos-ui/svelte-headless/slot';
+
+	// cf https://github.com/ota-meshi/eslint-plugin-svelte/issues/348
```
MarkoOleksiyenko, line 5: same here: comment not needed as it is used a bit after

angular/lib/src/components/slider/slider.component.ts
andreascorti: Done, had to create also the corresponding `AdaptSlotContentProps` to match interfaces

angular/demo/src/app/samples/slider/custom.route.ts
```diff
@@ -0,0 +1,74 @@
...
+					[attr.aria-disabled]="state.disabled ? true : null"
+					[attr.aria-valuenow]="item.value"
+					[attr.aria-valuetext]="item.value"
+					[attr.aria-label]="item.ariaLabel"
+					[attr.aria-orientation]="state.vertical ? 'vertical' : null"
+					[disabled]="state.disabled"
+					[class]="state.vertical ? 'au-custom-slider-handle-vertical' : 'au-custom-slider-handle-horizontal'"
+					[style.left.%]="state.handleDisplayOptions[item.id].left"
+					[style.top.%]="state.handleDisplayOptions[item.id].top"
+					(keydown)="onKeyDown($event, item.id, widget.actions.keydown)"
```
MarkoOleksiyenko, line 34: So the problem of the focus comes from the fact that the contents of the button is an svg, so the _event.target_ is a path of said svg. We can avoid having the 'custom' methods on the button, if you modify the **mouseDown** core method to use _event.currentTarget_ instead of the _event.target_. It takes the actual element on which we put event and not its contents (so in our case it will be button and you can remove _focusParent_ too). 
Be careful as apparently the currentTarget needs to be assigned straight away to a variable in order to be used, so in mouseDown you need to write: const currentTarget = event.currentTarget immediately, otherwise it will be null :)

angular/demo/src/app/samples/slider/fullcustom.route.ts
```diff
@@ -0,0 +1,64 @@
...
+	`,
+	styles: "@import '@agnos-ui/common/samples/slider/custom.scss';",
+	encapsulation: ViewEncapsulation.None,
+})
+export default class CustomSliderComponent {
+	private _zone = inject(NgZone);
+
+	sliderControl = new FormControl([50]);
+
+	onKeyDown(event: KeyboardEvent, handleId: number, widgetOnKeyDownFn: (event: KeyboardEvent, handleId: number) => void) {
```
MarkoOleksiyenko, line 58: same a for the custom.route.ts

angular/demo/src/app/samples/slider/custom.route.ts
andreascorti: That's a good suggestion that actually would make it works with any content of the button.
done

svelte/lib/src/components/slider/SliderDefaultHandle.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+	import type {SliderSlots} from '@agnos-ui/svelte-headless/components/slider';
+
+	type $$Props = SliderSlots['handle']; // eslint-disable-line @typescript-eslint/no-unused-vars
```
MarkoOleksiyenko, line 4: you can even remove the one after props I believe

svelte/lib/src/components/slider/SliderDefaultStructure.svelte
```diff
@@ -0,0 +1,128 @@
+<script lang="ts">
+	import type {SliderSlots} from '@agnos-ui/svelte-headless/components/slider';
+	import {Slot} from '@agnos-ui/svelte-headless/slot';
+
+	type $$Props = SliderSlots['structure']; // eslint-disable-line @typescript-eslint/no-unused-vars
```
MarkoOleksiyenko, line 5: and this comment too (minor but better to keep samples as clean as possible :))

core/src/components/slider/slider.spec.ts
```diff
@@ -66,25 +68,31 @@ describe(`Slider basic`, () => {
 		const sliderDirective = slider.directives.sliderDirective(sliderElement);
 		const minLabelDirective = slider.directives.minLabelDirective(minLabel);
 		const maxLabelDirective = slider.directives.maxLabelDirective(maxLabel);
-		const unsubscribe = slider.state$.subscribe((newState) => {
-			state = newState;
+
+		normalizedState$ = computed(() => {
```
MarkoOleksiyenko, line 72: mb we should extract the method to get the normalize state in case we have some other fields?

react/lib/src/components/slider/slider.tsx
```diff
@@ -1,51 +1,48 @@
...
-			showValueLabels,
-			showMinMaxLabels,
-			rtl,
-		},
-		{
-			actions: {click, keydown, mouseDown, touchStart},
-			directives: {sliderDirective, minLabelDirective, maxLabelDirective},
-		},
-	] = useWidgetWithConfig(createSlider, props, 'slider');
+function DefaultSlotHandle(slotContext: SliderSlotHandleContext) {
```
quentinderoubaix, line 12: the default Slots implementations should also be exported

svelte/demo/src/app/samples/slider/Custom.route.svelte
```diff
@@ -0,0 +1,51 @@
...
+	import '@agnos-ui/common/samples/slider/custom.scss';
+</script>
+
+<h2>Slider with custom labels</h2>
+<Slider
+	min={0}
+	max={100}
+	stepSize={1}
+	bind:values={sliderCustomLabelControl}
+	showMinMaxLabels={true}
```
quentinderoubaix, line 16: same thing, no need for showMin and showValue

svelte/demo/src/app/samples/slider/Custom.route.svelte
```diff
@@ -0,0 +1,51 @@
...
+	showValueLabels={true}
+	slotLabel={({value}) => '' + value + '%'}
+></Slider>
+<hr />
+<h2>Slider with custom handle</h2>
+<Slider min={0} max={100} stepSize={1} bind:values={sliderCustomHandleControl} showMinMaxLabels={true} showValueLabels={true}>
+	<svelte:fragment slot="handle" let:state let:item let:widget>
```
quentinderoubaix, line 23: The custom handle slot could also be exported in its own file.

svelte/demo/src/app/samples/slider/FullCustom.route.svelte
```diff
@@ -0,0 +1,55 @@
...
+<h2>Fully custom slider</h2>
+<Slider
+	min={0}
+	max={100}
+	stepSize={1}
+	bind:values={sliderControl}
+	showMinMaxLabels={true}
+	showValueLabels={true}
+	vertical={true}
+	className={'custom-slider'}
```
quentinderoubaix, line 18: could be just className="custom-slider"

react/demo/src/app/samples/slider/FullCustom.route.tsx
```diff
@@ -0,0 +1,71 @@
...
+				min={0}
+				max={100}
+				stepSize={1}
+				values={sliderControl}
+				onValuesChange={setSliderControl}
+				showMinMaxLabels={true}
+				showValueLabels={true}
+				slotStructure={CustomComponent}
+				vertical={true}
+				className={'custom-slider'}
```
quentinderoubaix, line 66: could be just className='custom-slider'

quentinderoubaix (commented):
thx for the PR ! @andreascorti 
left a few comments.

one last for Angular:
all directives should be exported in the global module (agnos-ui-angular.module.ts)

core/src/components/slider/slider.spec.ts
```diff
@@ -66,25 +77,25 @@ describe(`Slider basic`, () => {
...
 
 	test(`should create the default configuration for the model`, () => {
-		expect(state).toStrictEqual(defaultStateValues);
+		expect(normalizedState$()).toStrictEqual(defaultStateValues());
 	});
 
 	test(`should recompute clean value if min has changed`, () => {
-		const expectedStateValue = {...defaultStateValues};
-		expect(state).toStrictEqual(expectedStateValue);
+		const expectedStateValue = defaultStateValues();
```
fbasso, line 97: Not directly linked to this PR, but can you change all the `StateValue` expressions (expectedStateValue, defaultStateValues) to `State` (expectedState, defaultState). These two expressions has almost the same meaning.

fbasso (changes requested):
Regarding the test, just a minor comment, otherwise it's fine

fbasso (approved):
Fine for me, regarding the test ;)

angular/demo/src/app/samples/slider/custom.route.ts
```diff
@@ -0,0 +1,34 @@
+import {SliderComponent, SliderHandleDirective, SliderLabelDirective} from '@agnos-ui/angular';
+import {Component} from '@angular/core';
+import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
+import CustomHandleSliderComponent from './custom-handle.component';
+
+@Component({
+	standalone: true,
+	imports: [SliderComponent, SliderLabelDirective, SliderHandleDirective, ReactiveFormsModule, FormsModule],
```
quentinderoubaix, line 8: The import of ```SliderHandleDirective``` can be removed.

quentinderoubaix (approved):
LGTM !
thx for the PR @andreascorti

PR #514, opened 2024-02-21
feat: set tansu as peer dependency

### LINKED ISSUE #458: Put Tansu as a peer dependency
To avoid duplication if a project already use Tansu.

quentinderoubaix:
closes #458


andreascorti (commented):
thanks @quentinderoubaix ! LGTM, but I see angular/demo has no dependency on tansu. Is it normal?

quentinderoubaix:

PR #518, opened 2024-02-22
feat(Accordion): use clear naming closes #160

### LINKED ISSUE #160: Improve the tsdoc about the accordion widget, change the `accordion collapse` expression to make it more easily understandable
Especially, the tsdoc currently mentions the [accordion collapse](https://github.com/AmadeusITGroup/AgnosUI/blob/v0.1.1/core/src/components/accordion/accordion.ts#L120) as a DOM element in the structure (and repeated several times in the doc).
As suggested by @ExFlo, maybe it should be called accordion body instead?

ValentinNelu:
closes #160

### DIFF
--- a/core/src/components/accordion/accordion.ts
+++ b/core/src/components/accordion/accordion.ts
@@ -42,7 +42,7 @@ function getItem(items: AccordionItemWidget[], itemId: string): AccordionItemWid
 
 export interface AccordionProps extends WidgetsCommonPropsAndState {
 	/**
-	 * If `true`, only one item at the time can stay open.
+	 * If `true`, only one accordion-item at the time can stay open.
 	 */
 	closeOthers: boolean;
 	/**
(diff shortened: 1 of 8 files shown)


core/src/components/accordion/accordion.ts
```diff
@@ -294,28 +296,29 @@ export interface AccordionItemCommonPropsAndState {
...
 	 */
 	itemHeaderClass: string;
 	/**
-	 * Classes to add on the accordion-item collapse DOM element.
+	 * CSS classes to add on the accordion-item collapse DOM element.
 	 */
 	itemButtonClass: string;
 	/**
-	 * Classes to add on the accordion-item collapse DOM element.
+	 * CSS classes to add on the accordion-item collapse DOM element.
```
divdavem, line 317: I think you should also update this description, which is the same as the previous one, which should also be updated.

ValentinNelu: I've updated the description for both (Accordion and Accordion Item), thx for the remark

quentinderoubaix (approved):
LGTM !

PR #526, opened 2024-02-26
feat(blog): Add Intl support for publish date

MarkoOleksiyenko:


demo/src/lib/layout/Published.svelte
```diff
@@ -0,0 +1,7 @@
+<script>
+	export let date;
+
+	date = Intl.DateTimeFormat('en-US', {dateStyle: 'full'}).format(new Date(date));
```
quentinderoubaix, line 4: maybe it should be:
```typescript
$: dateString = Intl.DateTimeFormat('en-US', {dateStyle: 'full'}).format(new Date(date));
```

quentinderoubaix (commented):
thx for the PR @MarkoOleksiyenko, left you a few minor comments

demo/src/lib/layout/Published.svelte
```diff
@@ -0,0 +1,7 @@
...
+</script>
+
+<strong class="d-flex mb-2">Published on {date}</strong>
```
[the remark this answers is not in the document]
MarkoOleksiyenko: without it the mb is not applied

demo/src/lib/layout/Published.svelte
MarkoOleksiyenko: was thinking about the reactivity dangle, but as it is rendered only once not sure that it is actually needed, wdyt?

quentinderoubaix: It's not really needed indeed... though it provides better code readability.

PR #530, opened 2024-02-26
feat: progressbar slot content renamed to slot structure

### LINKED ISSUE #509: [progressbar] Rename slotContent to slotStructure
**Description**
Progressbar is the only widget that uses **slotContent** instead of **slotStructure** as the higher level slot.
This issue is to fix that inconsistency.

### LINKED ISSUE #524: [Progressbar] [Angular] AfterContentChecked uses wrong patch method
**Describe the bug**
AfterContentChecked should use `patchSlots` and not `patch`.

**To Reproduce**
Try to use the input for a slot structure in the progress bar. 
Customization won't be applied because `patch` is resetting the slot to `undefined`

**Expected behavior**
`patchSlots` does not reset the template, custom template is applied.

**Context**
The issue was spot by the implementation of slots in slider, PR https://github.com/AmadeusITGroup/AgnosUI/pull/507

quentinderoubaix:
closes #524 
closes #509


andreascorti (approved):
LGTM tnx @quentinderoubaix !

PR #544, opened 2024-02-28
test: Add e2e state to alert and modal

MarkoOleksiyenko:


e2e/alert/alert.e2e-spec.ts
```diff
@@ -1,6 +1,10 @@
 import {AlertPO} from '@agnos-ui/page-objects';
 import {expect, test} from '../fixture';
 import {AlertDemoPO} from '../demo-po/alert.po';
+import {assign} from 'common/utils';
+
+type PromiseValue<T> = T extends Promise<infer U> ? U : never;
```
ExFlo, line 6: should it not be a common type ?

e2e/alert/alert.e2e-spec.ts
```diff
@@ -10,20 +14,28 @@ test.describe(`Alert tests`, () => {
...
+
 		await alertDemoPO.locatorTypeSelect.selectOption('danger');
 		await alertDemoPO.locatorShowAlertButton.click();
 
-		await expect(alertPO.locatorRoot).toHaveClass(/alert-danger/);
-		await expect(alertPO.locatorRoot).not.toHaveClass(/alert-dismissible/);
-		await expect(alertPO.locatorCloseButton).toHaveCount(0);
+		expect(await alertPO.state()).toStrictEqual(
+			assign(expectedState, {
+				...expectedState,
```
fbasso, line 34: You don't need to add this line, this is already done in `assign`

page-objects/lib/alert.po.ts
```diff
@@ -16,4 +17,19 @@ export class AlertPO extends BasePO {
 	get locatorCloseButton(): Locator {
 		return this.locatorRoot.locator(this.selectors.closeButton);
 	}
+
+	async state() {
+		return this.locatorRoot.evaluate(async (rootNode: HTMLElement, selectors) => {
+			await Promise.all(rootNode.getAnimations().map((animation) => animation.finished));
```
fbasso, line 23: I didn't know this api, but are you sure it's usefull in our case too ? I'm curious to know if it also works with transitions (what we use), or only with animations...

e2e/modal/modal.e2e-spec.ts
```diff
@@ -75,4 +92,71 @@ test.describe.parallel(`Modal tests`, () => {
...
+		await test.step(`Modal with only body`, async () => {
+			await page.goto('#/modal/playground#{"props":{"visible":true,"slotDefault":"Body"}}');
+			expect(await modalPO.state()).toStrictEqual(expectedState);
+		});
+
+		await test.step(`Modal with body and footer`, async () => {
+			await page.goto('#/modal/playground#{"props":{"visible":true,"slotDefault":"Body","slotFooter":"Footer"}}');
+			expect(await modalPO.state()).toStrictEqual(
+				assign(expectedState, {
+					...expectedState,
```
fbasso, line 116: Not needed (same for other steps)

e2e/modal/modal.e2e-spec.ts
```diff
@@ -75,4 +92,71 @@ test.describe.parallel(`Modal tests`, () => {
...
+		const expectedState: State = {
+			rootClasses: ['d-block', 'fade', 'modal', 'show'],
+			header: undefined,
+			title: undefined,
+			body: 'Body',
+			footer: undefined,
+			closeButton: undefined,
+		};
+
+		await test.step(`Modal with only body`, async () => {
```
fbasso, line 107: Instead of doing steps, I thing it would be better to have a normal flow with a `describe` and separated `test`. The point here is that each test are dependant on the previous one. At the end, it can be difficult to quickly understand how is built the expected states through the different steps. 

Also, maybe it would be a good idea to have a function that returns the expectedState, so that a new object will be created each time `getExpectedState` will be called, to ensure to work with an independant object all the time.

fbasso (changes requested):
Thanks for you PR @MarkoOleksiyenko ! It's good overall, just some little changes to do/discuss.

page-objects/lib/alert.po.ts
MarkoOleksiyenko: in fact after we have modified the 'fade' animation it is not needed anymore as fade always stays. The issue was that the state was returned before the fade was removed, so we needed animation.finished.
Don't know whether it works with transitions though - it is not specified in https://developer.mozilla.org/en-US/docs/Web/API/Element/getAnimations

e2e/modal/modal.e2e-spec.ts
```diff
@@ -75,4 +92,76 @@ test.describe.parallel(`Modal tests`, () => {
 		await modalPO.locatorCloseButton.click();
 		await modalPO.locatorRoot.waitFor({state: 'hidden'});
 	});
+
+	test.describe.parallel(`Modal customizability`, async () => {
+		const getExpectedState: () => State = () => {
```
fbasso, line 97: You can simplify the code with :

```ts
    const getExpectedState = (): State => ({
        rootClasses: ['d-block', 'fade', 'modal', 'show'],
        header: undefined,
        title: undefined,
        body: 'Body',
        footer: undefined,
        closeButton: undefined,
    });
```

page-objects/lib/alert.po.ts
```diff
@@ -16,4 +17,18 @@ export class AlertPO extends BasePO {
 	get locatorCloseButton(): Locator {
 		return this.locatorRoot.locator(this.selectors.closeButton);
 	}
+
+	async state() {
+		return this.locatorRoot.evaluate(async (rootNode: HTMLElement, selectors) => {
```
fbasso, line 22: No need to be async anymore

fbasso (changes requested):
Thanks for your changes ! Just some (new) minor ones...

fbasso (approved):
Fine for me, thanks @MarkoOleksiyenko ;)

PR #546, opened 2024-02-28
fix: running local after a build

quentinderoubaix:
We have a issue currently that can be reproduced by the following steps:

Run
```bash
git clean -dfX && npm ci && npm run build:ci && npm run dev
```
Observe the issue, delete the dist folder in the core folder, run ```npm run dev``` and observe that it's back to normal.

This PR solves this.


andreascorti (approved):
LGTM @quentinderoubaix nice, thanks!

PR #559, opened 2024-03-05
fix: transition demos should use useMemo instead of useWidget

quentinderoubaix:


ExFlo (approved):

quentinderoubaix:
thx for the review @ExFlo !

PR #561, opened 2024-03-06
ci(angular): fail build when an angular directive or component is not exported

### LINKED ISSUE #548: Make sure the build fails if an Angular directive or component is not exported in the main module
**Description**
We tend to forget to export the newly created directives / components in the main angular module, which should be mandatory.
In order to avoid this, we should either make this angular module generated automatically or add a script that will fail the ci if it is missing.

quentinderoubaix:
closes #548

### DIFF
--- a/angular/lib/scripts/verify-module.mjs
+++ b/angular/lib/scripts/verify-module.mjs
@@ -1,0 +1,50 @@
+import {glob} from 'glob';
+import path from 'path';
+import {fileURLToPath} from 'url';
+import {readFile} from 'fs/promises';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const pathRegExp = /\\/g;
+function normalizePath(str) {
+	return str.replace(pathRegExp, '/');
+}
+
+const COMPONENTS_FOLDER = path.join(__dirname, '../src/components');
+
+const files = (
+	await glob('**/*.component.ts', {
+		cwd: COMPONENTS_FOLDER,
+		nodir: true,
+	})
+).map((file) => normalizePath(path.join(COMPONENTS_FOLDER, file)));
+
+const stuffToExportRegex = /(@Directive|@Component)\({[\S\s]+?}\)\s*export class ([a-zA-Z]*)/g;
+
+const classesToExport = [];
+for (const file of files) {
+	const content = await readFile(file, {encoding: 'utf-8'});
+	const matches = [...content.matchAll(stuffToExportRegex)];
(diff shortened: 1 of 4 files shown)


angular/lib/scripts/verify-module.mjs
```diff
@@ -0,0 +1,48 @@
+import {glob} from 'glob';
+import path from 'path';
+import {fileURLToPath} from 'url';
+import {readFile, writeFile, mkdir, rm} from 'fs/promises';
```
MarkoOleksiyenko, line 4: writeFile, mkdir, rm are not used

angular/lib/scripts/verify-module.mjs
```diff
@@ -0,0 +1,48 @@
...
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const pathRegExp = /\\/g;
+export function normalizePath(str) {
```
MarkoOleksiyenko, line 9: do we need to export the function here?

angular/lib/scripts/verify-module.mjs
```diff
@@ -0,0 +1,48 @@
...
+	return str.replace(pathRegExp, '/');
+}
+
+const files = (
+	await glob('**/*.component.ts', {
+		cwd: path.join(__dirname, '../src/components'),
```
MarkoOleksiyenko, line 15: minor: i feel every time that the string is repeated 2+ times it is better to create a var for it (../src/components)

angular/lib/scripts/verify-module.mjs
```diff
@@ -0,0 +1,50 @@
...
+const COMPONENTS_FOLDER = path.join(__dirname, '../src/components');
+
+const files = (
+	await glob('**/*.component.ts', {
+		cwd: COMPONENTS_FOLDER,
+		nodir: true,
+	})
+).map((file) => normalizePath(path.join(COMPONENTS_FOLDER, file)));
+
+const stuffToExportRegex = /(@Directive|@Component)\({[\S\s]+?}\)\s*export class ([a-zA-Z]*)/g;
```
andreascorti, line 22: We have a use case of "empty" `@Directives()` in the headless, but I think your regex is more than enough for components (in general you have at least the `selector`).

MarkoOleksiyenko (approved):
lgtm, thanks for the PR @quentinderoubaix

andreascorti (approved):
LGTM, nice way to prevent these issues @quentinderoubaix !

PR #566, opened 2024-03-07
feat: angular decorators to signal functions - progress bar component example 

mrednic-1A:


quentinderoubaix:
thanks for the initial PR @mrednic-1A !
based on your initial work, we are handling the migration to signals in #1029

PR #567, opened 2024-03-07
feat(demo): Add portal version selection dropdown

MarkoOleksiyenko:


demo/src/routes/menu/Versions.svelte
```diff
@@ -0,0 +1,86 @@
...
+
+	interface Version {
+		folder: string;
+		version: string;
+	}
+
+	const versionsParsed: Version[] = [{folder: 'latest', version: 'Latest'}];
+
+	onMount(() => {
+		fetch('https://amadeusitgroup.github.io/AgnosUI/versions.json').then((res) => {
```
MarkoOleksiyenko, line 18: was thinking to add the pathToRoot/versions.json but it won't work on local (kinda useless but still interesting to have). WDYT?

demo/src/routes/menu/Versions.svelte
```diff
@@ -0,0 +1,86 @@
...
+			res.json().then((result) => {
+				versionsParsed.push(...result);
+			});
+		});
+		({hasFocus$, directive} = createHasFocus());
+	});
+
+	const currentVersion$: Writable<Version> = writable(versionsParsed[0]);
```
quentinderoubaix, line 26: currentVersion initial value should probably be built using AGNOSUI_VERSION, something like this:
```typescript
  const currentVersion$ = writable(versions.find((version) => version.version === `v${import.meta.env.AGNOSUI_VERSION}`) ?? versions[0]);
```

quentinderoubaix:
Instead of building the versions in the onMount of the Versions component, you could do like this:

**+layout.ts**
```typescript
export const prerender = true;

export const load = async () => {
    const versionFetch = await fetch('https://amadeusitgroup.github.io/AgnosUI/versions.json');
    const versions = [{folder: 'main', version: 'PREVIEW'}].concat(await versionFetch.json());
    versions[1].folder === 'latest';
    return {
        versions
    };
}
```

**+layout.svelte**
```svelte
<script>
   export let data;
</script>
...
<Versions versions={data.versions} />
```

**Versions.svelte**
```svelte
<script>
  interface Version {
    folder: string;
    version: string;
  }
  export let versions: Version[];
</script>
```

demo/src/routes/menu/Versions.svelte
```diff
@@ -0,0 +1,86 @@
...
+
+	function giveFocus(el: HTMLAnchorElement, index: number) {
+		if (index === 0) {
+			el.focus();
+		}
+	}
+
+	function getRedirectionURL(version: string) {
+		const currentURL = get(page).url;
+		if (get(page).data['since'] <= version) {
```
quentinderoubaix, line 41: You should probably use lte from semver to compare versions instead of `<=`

demo/src/routes/menu/Versions.svelte
```diff
@@ -0,0 +1,86 @@
...
+	let directive: Directive;
+	$: $open$ = $hasFocus$;
+
+	function giveFocus(el: HTMLAnchorElement, index: number) {
+		if (index === 0) {
+			el.focus();
+		}
+	}
+
+	function getRedirectionURL(version: string) {
```
quentinderoubaix, line 39: Instead of having this function, you could create a reactive list depending on versions and the page store.
Something like this:
```svelte
<script>
  $: versionsWithUrl = versions.map((version) => ({
    ...version,
    url:
      version.version === 'PREVIEW' || ($page.data['since'] && valid($page.data['since']) && lte($page.data['since'], version.version))
        ? $page.url.pathname.replace(`/${$currentVersion$.folder}/`, `/${version.folder}/`)
        : `${$pathToRoot$.replace($currentVersion$.folder, version.folder)}docs/${$selectedFramework$}/getting-started/introduction`,
  }));
</script>
```

demo/src/routes/menu/Versions.svelte
```diff
@@ -0,0 +1,86 @@
...
+		} else {
+			return `${$pathToRoot$.replace($currentVersion$.folder, version)}docs/${$selectedFramework$}/getting-started/introduction`;
+		}
+	}
+</script>
+
+<div class="nav-item ms-3">
+	<div class="dropdown">
+		<button
+			class="btn nav-link dropdown-toggle align-items-center d-flex"
```
quentinderoubaix, line 52: If you add the PREVIEW, targetting main, it could be good to change the styling of the button to make it clear to users it's not released yet.
Maybe something like this ?
```svelte
<button
  class="btn nav-link dropdown-toggle align-items-center d-flex {$currentVersion$.version === 'PREVIEW' ? 'badge text-bg-warning' : ''}"
 ></button>
```

quentinderoubaix (approved):
LGTM !

PR #578, opened 2024-03-12
fix: type guard on core component stores

### LINKED ISSUE #511: `State` interfaces are not granted by `stateStores` function
**Describe the bug**
Currently the interfaces we are defining for States are not checked at build time in the return state of widgets.
The function `stateStores` accepts stores that are not defined in the `State` interface, and it is not failing at build time. 

Right now it is possible to spot it in unit tests, with a strict equal on the state.

**To Reproduce**

1) Add an extra store to any `stateStores` without defining it in the corresponding state of a component.
2) Build does not raise any error.
3) Unit test on expected state fails if the expected state does not include the new property.

**Expected behavior**
We should check if it is possible to verify that no additional properties are added compared to the original `State` defined by the widget.

quentinderoubaix:
closes #511


andreascorti (approved):
LGTM, thanks @quentinderoubaix

quentinderoubaix:

PR #581, opened 2024-03-12
fix(build): missing dependency on parent tsconfig.json files

divdavem:


quentinderoubaix:
Maybe we should also add the tsconfig for the angular and react workspaces ?

quentinderoubaix (approved):
LGTM !

divdavem:

PR #584, opened 2024-03-12
fix(focustrack): avoid failing when reading activeElement$ server-side

divdavem:


andreascorti (approved):

divdavem:
@quentinderoubaix @andreascorti Thank you for your review!

PR #592, opened 2024-03-14
Workaround for SSR svelte issue with slots

divdavem:
This PR reverts #522 and implements the workaround mentioned in https://github.com/sveltejs/svelte/issues/9137, which is replacing:
```svelte
<MyComponent>
<slot name="otherComponentSlot" slot="myComponentSlot" let:props {...props}/>
</MyComponent>
```
by:
```svelte
<MyComponent>
<svelte:fragment slot="myComponentSlot" let:props><slot name="otherComponentSlot" {...props}/></svelte:fragment>
</MyComponent>
```


quentinderoubaix (approved):
thx for the PR @divdavem
can't wait to use Svelte 5 to make all of this code far simpler !

PR #604, opened 2024-03-18
fix(build): make build:src a dependency of build:pkg in core and react

divdavem:


quentinderoubaix (approved):
thx for the fix @divdavem !

divdavem:

PR #610, opened 2024-03-19
fix(lint): also lints js files

divdavem:


quentinderoubaix (approved):

divdavem:

PR #613, opened 2024-03-19
fix(core): fixing core mergeInto function

divdavem:


quentinderoubaix (approved):
thx for the fix @divdavem !

PR #624, opened 2024-03-20
ci: boost demo build

quentinderoubaix:


demo/package.json
```diff
@@ -91,8 +91,8 @@
 		"@sveltejs/vite-plugin-svelte": "^3.0.2",
 		"@types/dom-view-transitions": "^1.0.4",
 		"bootstrap-icons": "^1.11.3",
+		"compare-versions": "^6.1.0",
```
andreascorti, line 94: Shouldn't we use it also in our scripts and remove completely the dependency to `semver` then?

andreascorti (commented):
Hey @quentinderoubaix, I don't understand why we use 2 libraries for the same thing (semver / compare-versions).

scripts/versionDocFolder.js
```diff
@@ -1,4 +1,4 @@
-const {parse} = require('semver');
+const regexMajorMinor = /^(\d+)\.(\d+)/;
```
andreascorti, line 1: had a look quickly to the library, shouldn't you use `validateAndParse`? https://github.com/omichelsen/compare-versions/blob/main/src/utils.ts#L9C14-L9C30

andreascorti: my bad, they do not export utils

andreascorti (approved):

quentinderoubaix:

PR #638, opened 2024-03-21
refactor(demo): refactor stackblitz to prepare for tests

divdavem:


demo/src/lib/stackblitz/prepareProject.ts
```diff
@@ -0,0 +1,90 @@
+import type {SampleInfo} from '$lib/layout/sample';
+import type {Frameworks} from '$lib/stores';
+import type {Project} from '@stackblitz/sdk';
+import type {StackblitzProcessor} from './utils';
+import {addAsyncFiles, isBootstrapCondition, isDaisyuiCondition} from './utils';
+
+export const createBaseFrameworkProcessors = (): Record<Frameworks, StackblitzProcessor[]> => ({
```
quentinderoubaix, line 7: Maybe move the `createBaseFrameworkProcessors` and `createFrameworkProcessorsReleased` somewhere else ?
It was surprising to see these functions in the prepareProject file instead of frameworkProcessors

demo/vite.config.ts
```diff
@@ -60,6 +61,7 @@ export default defineConfig((config) => ({
 	plugins: [
 		copySamples(),
 		includeSamples(),
+		listSamples(),
```
quentinderoubaix, line 64: We don't need the plugin in main vite.config for the moment (it has zero impact, but just to be cleaner).

quentinderoubaix (approved):
Thanks for the PR !
Left some very very minor comments.

demo/src/lib/stackblitz/prepareProject.ts
divdavem: As discussed, I renamed the file `unreleasedProcessors`.

demo/vite.config.ts
divdavem: Indeed it is not needed in the main vite.config. I have removed it.

divdavem:

PR #647, opened 2024-03-22
feat: ssr features for directives

fbasso:


core/src/utils/stores.ts
```diff
@@ -141,6 +141,13 @@ export const toReadableStore = <T>(x: ReadableSignal<T> | T) => (isStore(x) ? x
  */
 export const toWritableStore = <T>(x: WritableSignal<T> | T) => (isStore(x) ? x : writable(x));
 
+/**
+ *  If the provided argument is a store, its value is returned, otherwise, the value is returned
```
divdavem, line 145:
```suggestion
 * If the provided argument is a store, its value is returned, otherwise, the value is returned
```
(there seems to be an extra space before `If the provided argument`)

react/headless/src/utils/directive.ts
```diff
@@ -49,3 +51,35 @@ export function useDirectives<T>(directives: Directive<T>[], args?: T): RefCallb
...
+	['for', 'htmlFor'],
+]);
+
+/**
+ * Returns an object with the key/value attributes for JSX, derived from a list of directives.
+ *
+ * @param directives - List of directives to generate attributes from. Each parameter can be the directive or an array with the directive and its parameter
+ * @returns JSON object with key/value pairs to be applied on a JSX node.
+ */
+export function ssrAttributes<T extends any[]>(...directives: {[K in keyof T]: AttributeDirectiveAndParam<T[K]> | AttributeDirective<void>}) {
```
divdavem, line 66: In the core, the `ssrAttributes` function is just a shortcut to the `directiveAttributes` function depending on the environment. It should be the same in react.

core/src/utils/directive.ts
```diff
@@ -227,76 +228,150 @@ export const mergeDirectives =
...
+ * Returns an object with the attributes, style and class keys containing information derived from a list of directives.
+ *
+ *   - The `attributes` value is a JSON representation of key/value attributes, excepted for the `class` and `style` attributes
+ *   - The `classNames` value is an array of string representing the classes to be applied
+ *   - The `style` value is a JSON representation of the styles to be applied
+ *
+ * @param directives - List of directives to generate attributes from. Each parameter can be the directive or an array with the directive and its parameter
+ * @returns JSON object with the `attributes`, `class` and `style` keys.
+ */
+export function ssrAttributesData<T extends any[]>(...directives: {[K in keyof T]: AttributeDirectiveAndParam<T[K]> | AttributeDirective<void>}) {
```
divdavem, line 303: Actually, even if this function is mostly intended for SSR, nothing prevents it from being used for something else.
Maybe we should reserve the `ssr` prefix for functions that just call the unprefixed function of the same name only when not in a browser environment.

fbasso:
Close in flavor of #683

PR #654, opened 2024-03-26
feat(slider): Refactor using attributes directives

andreascorti:
- [X] Create a directive to manage handle
- [x] Refactor container directive 
- [x] Progress display option directive?
- [x] Min Label directive including aria / class  & merge existing directive
- [x] Max label directive including aria / class & merge existing  directive?
- [x] Handle labels directives (combined + standard)

### DIFF
--- a/core/src/components/slider/slider.ts
+++ b/core/src/components/slider/slider.ts
@@ -1,7 +1,7 @@
-import type {WritableSignal} from '@amadeus-it-group/tansu';
+import type {ReadableSignal, WritableSignal} from '@amadeus-it-group/tansu';
 import {computed, writable} from '@amadeus-it-group/tansu';
 import type {WidgetsCommonPropsAndState} from '../commonProps';
-import {createStoreDirective, mergeDirectives} from '../../utils/directive';
+import {createAttributesDirective, createStoreDirective, mergeDirectives} from '../../utils/directive';
 import type {ConfigValidator, Directive, PropsConfig, SlotContent, Widget, WidgetSlotContext} from '../../types';
 import {noop} from '../../utils/internal/func';
 import {getDecimalPrecision} from '../../utils/internal/math';
@@ -224,6 +224,21 @@ export interface SliderDirectives {
 	sliderDirective: Directive;
 
 	/**
+	 * Directive used to style the progress display for each handle
+	 */
+	progressDisplayDirective: Directive<{option: ProgressDisplayOptions}>;
+
+	/**
+	 * Directive to apply to the slider clickable area, to directly move the handle to a given specific position
+	 */
+	clickableAreaDirective: Directive;
+
+	/**
+	 * Directive to apply to the slider handle if any
+	 */
+	handleDirective: Directive<{item: SliderHandle}>;
+
+	/**
 	 * Directive to get the minLabel elementRef
 	 */
 	minLabelDirective: Directive;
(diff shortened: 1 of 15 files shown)


core/src/components/slider/slider.ts
```diff
@@ -624,6 +629,32 @@ export function createSlider(config?: PropsConfig<SliderProps>): SliderWidget {
...
+					'aria-disabled': disabled$() ? 'true' : undefined,
+					disabled: disabled$() ? 'true' : undefined,
+					'aria-readonly': readonly$() ? 'true' : undefined,
+				},
+				styles: {
+					left: computed(() => handleDisplayOptions$()[handleContext$().item.id].left + '%'),
+					top: computed(() => handleDisplayOptions$()[handleContext$().item.id].top + '%'),
+				},
+				classNames: {
+					//TODO: add classes once directives can be customized.
```
andreascorti, line 655: I couldn't put `classNames` because otherwise they will be there even if in the custom handles I want them to be replaced by custom classes. It is an example of how important is the customization option.

svelte/lib/src/components/slider/SliderDefaultStructure.svelte
```diff
@@ -33,11 +33,11 @@
 		use:widget.directives.minLabelDirective
```
andreascorti, line 33: This section is also interesting: someone creating a slider with headless would probably have a different DOM, does it make sense to create directives for all those sub-elements ?

core/src/components/slider/slider.ts
fbasso: It's true that it's going to add classes, but I would say that it's not a problem as here, it's just specific classes from AgnosUI, not bootstrap. So I think it's fine to manage them from here, it's just classes that identify this specific element. We can have the same question for the other attributes.

But you're right, we'll need a way to remove them if a dev absolutely want to do so, and it will be part of the customization, that will be implemented later (the classes will have the same possibility of customization as the event, attributes an styles).

So, to be discussed, but my thought is that you can add them (only bootstrap specific must be avoided).

andreascorti: Ok, I only had to override bootstrap `_slider.scss` classes to get the custom styles I wanted.
A user of headless component won't have any css included by default, so it will be up to him to add css to the proper classes added by default by our API

core/src/components/slider/slider.ts
```diff
@@ -623,9 +685,83 @@ export function createSlider(config?: PropsConfig<SliderProps>): SliderWidget {
...
+				attributes: {
+					role: 'slider',
+					'aria-valuemin': min$,
+					'aria-valuemax': max$,
+					'aria-valuenow': computed(() => handleContext$().item.value),
+					'aria-valuetext': computed(() => handleContext$().item.ariaValueText),
+					'aria-label': computed(() => handleContext$().item.ariaLabel),
+					'aria-orientation': computed(() => (vertical$() ? 'vertical' : undefined)),
+					'aria-disabled': computed(() => (disabled$() ? 'true' : undefined)),
+					disabled: computed(() => (disabled$() ? 'true' : undefined)),
```
fbasso, line 726: This must be `disabled: disabled$`. It is not like aria-disabled where "true" (as a string) is expected. This should explain the change in bootstrap-slider-default.html.

```suggestion
					disabled: disabled$,
```

fbasso (approved):
Fine for me !

PR #661, opened 2024-03-27
feat: esm scripts + import.meta.dirname

quentinderoubaix:


divdavem:
Thank you for your PR!
I am fine with most of it (converting commonjs to ecmascript modules), but, as discussed, please do not convert typescript files to javascript (in `scripts/doc` folder), because it is less powerful and less strict for types, and not automatically type-checked anymore.

divdavem:
Thank you @quentinderoubaix 
Now that you switched from `ts-node` which does types validation to `tsx` which does not do type validation, you should make sure types are validated at some point in the CI.

quentinderoubaix:
> Thank you @quentinderoubaix Now that you switched from `ts-node` which does types validation to `tsx` which does not do type validation, you should make sure types are validated at some point in the CI.

@divdavem indeed, thanks for catching that !
I added a tsc check

scripts/run.js
```diff
@@ -1,21 +1,26 @@
-const path = require('path');
+#!/usr/bin/env -S node --no-warnings=ExperimentalWarning
```
divdavem, line 1: This line is useless, unless `run.js` is called as an executable, which is not the case in our setup, especially when called from `npm run`, where we specify explicitly `node scripts/run.js`. I think we should remove this line.

divdavem (approved):

quentinderoubaix:

PR #663, opened 2024-03-27
refactor(test): e2e test configuration and launch script

divdavem:


quentinderoubaix:
The command `npm run e2e chromium react rating -- --coverage` is failing as the demo workspace is missing the command `dev:coverage`

quentinderoubaix:
> The command `npm run e2e chromium react rating -- --coverage` is failing as the demo workspace is missing the command `dev:coverage`

It probably should not need the demo to be running either right ?

divdavem:
> > The command `npm run e2e chromium react rating -- --coverage` is failing as the demo workspace is missing the command `dev:coverage`
> 
> It probably should not need the demo to be running either right ?

Actually, the meaning of the command changed a bit compared to what was done previously. The rating test is looked for in all projects, including the demo project, that's why the demo is started.

quentinderoubaix (approved):
LGTM !

quentinderoubaix:
Maybe we should add a few lines in DEVELOPER.md to explain the difference of main vs singleBrowser vs demo.

divdavem:

PR #664, opened 2024-03-28
feat(accordion): directive implementation

### LINKED ISSUE #630: Refactor accordion to use core directives
**Description**
Following the work done on #539, this issue will track the refactor of the accordion to use agnostic core directives.

ValentinNelu:
closes #630

### DIFF
--- a/angular/demo/bootstrap/src/app/samples/accordion/customHeaders.route.ts
+++ b/angular/demo/bootstrap/src/app/samples/accordion/customHeaders.route.ts
@@ -1,16 +1,17 @@
-import {AgnosUIAngularModule} from '@agnos-ui/angular';
+import {AgnosUIAngularModule, UseDirective} from '@agnos-ui/angular';
 import {Component} from '@angular/core';
 import BODY from '@agnos-ui/common/samples/accordion/body.txt';
 
 @Component({
 	standalone: true,
-	imports: [AgnosUIAngularModule],
+	imports: [AgnosUIAngularModule, UseDirective],
 	template: `
 		<div auAccordion #accordion="auAccordion">
 			<div auAccordionItem>
 				<ng-template auAccordionItemStructure let-state="state" let-widget="widget">
 					<div
-						class="{{ state.itemHeaderClass }} accordion-button accordion-header custom-header justify-content-between"
+						[auUse]="widget.directives.headerDirective"
+						class="accordion-button accordion-header custom-header justify-content-between"
 						[class.collapsed]="!state.itemVisible"
 						role="heading"
 						aria-level="2"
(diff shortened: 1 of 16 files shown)


angular/demo/bootstrap/src/app/samples/accordion/customHeaders.route.ts
```diff
@@ -1,43 +1,34 @@
...
 					>
 						<p class="m-0">First panel - {{ state.itemVisible ? 'opened' : 'collapsed' }}</p>
 						<button
 							type="button"
-							id="{{ state.itemId }}-toggle"
-							(click)="widget.actions.click()"
-							class="btn btn-link p-0 {{ state.itemButtonClass }}"
-							[class.collapsed]="!state.itemVisible"
+							class="btn btn-link p-0 {{ state.itemButtonClass }} au-accordion-item-button"
 							[disabled]="state.itemDisabled"
```
quentinderoubaix, line 23: Why is the *disabled* attribute not managed by the toggleDirective ?

core/src/components/accordion/accordion.ts
```diff
@@ -493,22 +516,37 @@ function createAccordionItem(
...
 	});
+	const clickAction = () => {
+		if (!itemDisabled$()) {
+			itemVisible$.update((c: boolean) => !c);
+		}
+	};
+	const toggleDirective = createAttributesDirective(() => ({
+		attributes: {
+			id: computed(() => `${itemId$()}-toggle`),
+			'aria-expanded': computed(() => `${itemVisible$()}`),
```
quentinderoubaix, line 527: Can be replaced to `'aria-expanded': itemVisible$,`

core/src/components/accordion/accordion.ts
```diff
@@ -524,7 +562,34 @@ function createAccordionItem(
...
+				createAttributesDirective(() => ({
+					attributes: {
+						id: computed(() => `${itemId$()}-body-container`),
+						class: itemBodyContainerClass$(),
+						'aria-labelledby': computed(() => `${itemId$()}-toggle`),
+					},
+					classNames: {'au-accordion-item-body-container': true},
+				})),
+			),
+			accordionItemDirective: noop,
```
quentinderoubaix, line 591: Should the *accordionItemDirective* not be created like it was done for the *starDirective* in the rating ?

angular/demo/bootstrap/src/app/samples/accordion/customHeaders.route.ts
ValentinNelu: because is part of the button directive, the idea is that you could use the toggle directive even on elements that are not buttons. It's coming from ng-bootstrap approach needed to showcase the custom header

core/src/components/accordion/accordion.ts
ValentinNelu: It's a different approach because part of the directive is created inside the accordion, not in the accordion item. I could have created the attribute directive and then merged it inside the accordion but I think it is clearer to handle both at the same time, in the accordion

quentinderoubaix (approved):
thanks !

PR #669, opened 2024-03-29
fix: run.js command on windows

quentinderoubaix:


MarkoOleksiyenko (approved):
works for me, thanks for the PR @quentinderoubaix !

PR #673, opened 2024-03-29
feat(directives): Add support for options in events listeners used by createAttributesDirective

### LINKED ISSUE #672: Add support for options of event listeners added by createAttributesDirective
Today `createAttributesDirective` events support only an event function to attach to a listener, but not options. 

**Describe the solution you'd like**
With this feature I propose to change this behavior and modify the type of events to:

`Partial<{[K in keyof HTMLElementEventMap]: {eventFn: (this: HTMLElement, event: HTMLElementEventMap[K]) => void, options:? boolean | AddEventListenerOptions}>;`

**Impacts**:
Components using events in the createAttributesDirective.

andreascorti:
closes #672


core/src/utils/internal/dom.ts
```diff
@@ -68,24 +68,32 @@ export const removeClasses = (element: HTMLElement, classes?: string[]) => {
...
-	element.addEventListener(type, fn);
+export function addEvent(
+	element: Element,
+	type: string,
+	fn: EventListenerOrEventListenerObject,
+	options?: boolean | AddEventListenerOptions,
+): () => void {
+	element.addEventListener(type, fn, options);
 	return function () {
 		element.removeEventListener(type, fn);
```
divdavem, line 98: You should also pass the options in `removeEventListener`

core/src/utils/directive.ts
```diff
@@ -235,7 +235,12 @@ export interface AttributesDirectiveProps {
 	 * @remarks
 	 * Key-value pairs where keys are event types and values are event handlers.
 	 */
-	events?: Partial<{[K in keyof HTMLElementEventMap]: (this: HTMLElement, event: HTMLElementEventMap[K]) => void}>;
+	events?: Partial<{
+		[K in keyof HTMLElementEventMap]: {
+			eventFn: (this: HTMLElement, event: HTMLElementEventMap[K]) => void;
```
divdavem, line 240:
```suggestion
			handler: (this: HTMLElement, event: HTMLElementEventMap[K]) => void;
```
I suggest to name it `handler` rather than `eventFn`.

core/src/utils/directive.ts
```diff
@@ -235,7 +235,12 @@ export interface AttributesDirectiveProps {
...
+			options?: boolean | AddEventListenerOptions;
+		};
+	}>;
```
divdavem, line 243: Should you also accept passing the event handler directly without an object, as it is a common case (and before your PR the only possible way) and is easier to read?

core/src/utils/internal/dom.ts
```diff
@@ -68,24 +68,32 @@ export const removeClasses = (element: HTMLElement, classes?: string[]) => {
...
+	options?: boolean | AddEventListenerOptions,
 ): () => void;
 
 /**
  * Adds an event listener to the specified element.
  *
  * @param element - The HTML element to which the event listener will be added.
  * @param type - A string representing the event type to listen for.
- * @param fn - The event listener function or object.
+ * @param fn - The event listener function or object.\
```
divdavem, line 86:
```suggestion
 * @param fn - The event listener function or object.
```

divdavem (changes requested):
I have suggested some changes. What do you think?

andreascorti:
Thanks @divdavem for the review. I modified a unit test to use the object and added a new unit test for the addEvent

divdavem (approved):
Thank you @andreascorti

PR #695, opened 2024-04-09
feat(matchMedia): add new service matchMedia

quentinderoubaix:


core/src/services/matchMedia.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Create a store tracking the state of a match media query.
+ *
+ * @param query the query to match
+ * @returns a readable store tracking the match media query state
+ */
+export const matchMedia = (query: string) =>
+	BROWSER
+		? readable(false, {
+				onUse({set}) {
+					const mql = window.matchMedia(query);
```
andreascorti, line 14: enhancement: I know the API has no validation, but won't be worth to add one on our side? or an eslint rule ?

quentinderoubaix: @andreascorti thanks for the proposal !
I have tried to come up with smth for the validation of the media query, when you have time to check

PR #697, opened 2024-04-09
refactor(react): avoid calling useDirective when element is not in the DOM

divdavem:
This PR refactors some react components to avoid calling `useDirective` when the element is not in the DOM.
This is important for SSR because soon (in a future PR), `useDirective` will call `ssrAttributes` and this will call the directive on the server, and this must be skipped if the element will not be in the DOM (as it is useless and because it can cause some errors).
As it is not allowed to call a hook a variable number of times in a component, some react components had to be split in smaller components.

Issue #315, opened 2023-12-13
Ensure the e2e tests fail if an error or warning is present in the console


quentinderoubaix:
**Is your feature request related to a problem? Please describe.**
To help properly validate our component implementations, we should make sure that we handle the errors / warnings raised by the frameworks.

**Describe the solution you'd like**
Update the e2e tests (make a fixture probably) to ensure no error / warning was raised during the execution.

divdavem:
Note that some warnings are only raised by frameworks in development mode. Maybe we should also run our e2e tests in dev mode?

PR #698, opened 2024-04-09
Server-side rendering with react

divdavem:
This PR is built on top of #697 
It adds support for server-side rendering with react.

### DIFF
--- a/core/src/components/accordion/accordion.ts
+++ b/core/src/components/accordion/accordion.ts
@@ -526,7 +526,7 @@ function createAccordionItem(
 			'aria-expanded': computed(() => `${itemVisible$()}`),
 			'aria-disabled': computed(() => `${itemDisabled$()}`),
 			'aria-controls': computed(() => `${itemId$()}-body-container`),
-			disabled: itemDisabled$(),
+			disabled: itemDisabled$,
 		},
 		classNames: {collapsed: computed(() => !itemVisible$())},
 		events: {click: clickAction},
--- a/react/bootstrap/src/components/accordion/accordion.tsx
+++ b/react/bootstrap/src/components/accordion/accordion.tsx
@@ -2,7 +2,7 @@ import {Slot} from '@agnos-ui/react-headless/slot';
 import type {ForwardRefExoticComponent, ForwardedRef, PropsWithChildren, RefAttributes} from 'react';
 import {createContext, forwardRef, useContext, useEffect, useImperativeHandle} from 'react';
 import {useWidgetWithConfig} from '../../config';
-import {useDirective} from '@agnos-ui/react-headless/utils/directive';
+import {useClassDirective, useDirectives} from '@agnos-ui/react-headless/utils/directive';
 import type {AdaptSlotContentProps, AdaptWidgetSlots, Directive, WidgetFactory, WidgetProps, WidgetState} from '@agnos-ui/react-headless/types';
 import type {AccordionItemApi} from '@agnos-ui/core-bootstrap/components/accordion';
 import {createAccordion as coreCreateAccordion} from '@agnos-ui/core-bootstrap/components/accordion';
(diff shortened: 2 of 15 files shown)


core/src/components/accordion/accordion.ts
```diff
@@ -526,7 +526,7 @@ function createAccordionItem(
 			'aria-expanded': computed(() => `${itemVisible$()}`),
 			'aria-disabled': computed(() => `${itemDisabled$()}`),
 			'aria-controls': computed(() => `${itemId$()}-body-container`),
-			disabled: itemDisabled$(),
+			disabled: itemDisabled$,
```
quentinderoubaix, line 529: Thanks for catching this !

react/bootstrap/src/components/accordion/accordion.tsx
```diff
@@ -24,34 +24,29 @@ type AccordionTag = `h${1 | 2 | 3 | 4 | 5 | 6}`;
...
-	);
-};
+const DefaultSlotStructure = (slotContext: AccordionItemContext) => (
+	<>
+		<Header directive={slotContext.widget.directives.headerDirective} headerTag={slotContext.state.itemHeadingTag}>
+			<button {...useDirectives([useClassDirective('accordion-button'), slotContext.widget.directives.buttonDirective])}>
+				<Slot slotContent={slotContext.state.slotItemHeader} props={slotContext}></Slot>
+			</button>
+		</Header>
+		{slotContext.state.shouldBeInDOM ? <ItemContent {...slotContext} /> : null}
```
quentinderoubaix, line 46: Can be just
```suggestion
		{slotContext.state.shouldBeInDOM && <ItemContent {...slotContext} />}
```

react/bootstrap/src/components/alert/alert.tsx
```diff
@@ -30,29 +30,30 @@ const defaultConfig: Partial<AlertProps> = {
...
-					className={`au-alert alert alert-${state.type} ${state.className} ${state.dismissible ? 'alert-dismissible' : ''}`}
-					role="alert"
-					{...refTransition}
-				>
-					<Slot slotContent={state.slotStructure} props={slotContext}></Slot>
-				</div>
-			)}
-		</>
-	);
+	return <>{state.hidden ? null : <AlertElement {...slotContext} />}</>;
```
quentinderoubaix, line 58: Can be just
```suggestion
	return <>{!state.hidden && <AlertElement {...slotContext} />}</>;
```

react/headless/src/utils/directive.ts
```diff
@@ -1,61 +1,59 @@
...
+};
 
 const attributesMap = new Map([
 	['tabindex', 'tabIndex'],
 	['for', 'htmlFor'],
 ]);
 
+// For boolean attributes, the presence of the attribute means true, but react wants a boolean value
+// cf the list in https://github.com/facebook/react/blob/bf40b024421a0e1f2f882fd7171ea39cd74c88df/packages/react-dom-bindings/src/client/ReactDOMComponent.js#L665
+const booleanAttributes = new Set([
```
quentinderoubaix, line 31: As discussed internally, this is not maintainable in the long-term.
Am ok to merge as-is, but please create a issue to track this, and hopefully later on we will have time for a future-proof solution (maybe even nothing to do, if React were to export this enum ?)

react/ssr-app/vite.config.ts
```diff
@@ -8,4 +8,8 @@ export default defineConfig((config) => ({
 	resolve: {
 		alias: config.mode === 'production' ? {} : alias,
 	},
+	define: {
+		// make sure we always have react warnings even on the CI with "npm run preview"
+		'process.env.NODE_ENV': '"development"',
```
quentinderoubaix, line 13: I would rather remove this line.
In our CI, we want the e2e test of the react ssr app to run targetting a normal ssr react app...
For example, do you know if this also affects React development double-rendering ?

quentinderoubaix (commented):
Some minor comments

react/ssr-app/vite.config.ts
divdavem: yes, I think this also affects React development double-rendering
I think it is important not to neglect the development mode, there are more checks than in production mode and it does not look good if using AgnosUI in development brings a lot of errors that we do not detect in our test because we test in production mode.

quentinderoubaix: This is an interesting topic for sure, which could also be applied to the normal e2e tests, not just the ssr !
I opened the issue #315 some time ago so that we could one day properly avoid any error / warning, but it's outside the scope of this ssr react PR.

For this PR, I would propose either only production or both modes (as most dev-mode warnings would not be discovered by us anyway).

react/headless/src/utils/directive.ts
divdavem: I have opened #714 to track this

divdavem:
@quentinderoubaix I have made the changes you requested. I am not at all satisfied with the reduced validation this implies, but if it is ok for you, we can merge it like this and improve things later.

react/headless/src/utils/directive.ts
```diff
@@ -87,3 +85,52 @@ export function directiveAttributes<T extends any[]>(...directives: {[K in keyof
...
+ *
+ * @param directive - the directive
+ * @param args - the args to pass to the directive
+ * @returns the ref callback
+ */
+export const useDirective: {
+	(directive: Directive): {ref: RefCallback<HTMLElement>};
+	<T>(directive: Directive<T>, args: T): {ref: RefCallback<HTMLElement>};
+} = BROWSER
+	? <T>(directive: Directive<T>, args?: T): {ref: RefCallback<HTMLElement>; suppressHydrationWarning: true} => {
```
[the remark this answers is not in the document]
divdavem: I have opened #715 to track this.

quentinderoubaix (approved):
LGTM !

divdavem:
@quentinderoubaix Thank you!

PR #716, opened 2024-04-19
feat: multiDirective and corresponding refactoring in react and Angular

divdavem:
This PR adds the `multiDirective` directive in the core that can take as a parameter an array of directives to apply, optionally with their parameters: `[[myDirectiveWithParam, myParam], myDirectiveWithoutParam]`.
Using this directive, it is easy to implement in all frameworks the support for multiple directives when support for only one directive is already implemented.
This PR changes the `useDirectives` hook in react to now use `multiDirective` instead of `mergeDirectives` (with a change in its signature).
This PR also changes the `[auUse]` directive in Angular to also use `multiDirective` and provide multi-directive support in Angular (this is a breaking change).


quentinderoubaix (approved):
LGTM !

divdavem:

PR #725, opened 2024-04-22
refactor: remove useless CachedProperty

divdavem:
This is a breaking change: this PR removes the `@CachedProperty` decorator because it was mainly used to merge directives in a widget before using them with `auUse` (cf [this commit](https://github.com/AmadeusITGroup/AgnosUI/commit/0308b290c763da65761a0157960ddc831dca3321)), which only supports a single directive. Now that `auUseMulti` is available to apply multiple directives without requiring to merge them before, `CachedProperty` is no longer needed most of the time.


quentinderoubaix (approved):

divdavem:

PR #728, opened 2024-04-23
fix: svelte types in package @agnos-ui/svelte-bootstrap

### LINKED ISSUE #708: Fix "svelte-check" command after upgrade of svelte2tsx
**Context**
Currently we are forced to set the version of svelte2tsx to **0.7.0**.
This was done in this commit: https://github.com/AmadeusITGroup/AgnosUI/commit/80ccf7c8695b0fa95c748450a86e52c9825e97be

**Description**
Investigate why the types are invalid, make an issue / contribute to svelte2tsx if needed, then revert the above mentioned commit.

quentinderoubaix:
closes #708

### DIFF
--- a/.syncpackrc.js
+++ b/.syncpackrc.js
@@ -7,11 +7,6 @@ export default {
 	semverRange: '^',
 	versionGroups: [
 		{
-			packages: ['**'],
-			dependencies: ['@sveltejs/package'],
-			isIgnored: true,
-		},
-		{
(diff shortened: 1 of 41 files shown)


divdavem (approved):
Thank you @quentinderoubaix for solving this issue to be able to upgrade svelte2tsx.
As discussed, it would be good to have an automatic way to check that everything is consistent.

quentinderoubaix:
fully agree about this need to check our delivered types...

PR #732, opened 2024-04-24
feat: move all slot props from core to core-bootstrap

quentinderoubaix:


core/src/components/accordion/accordion.ts
```diff
@@ -589,122 +543,162 @@ function createAccordionItem(
...
+				},
+				expandAll: () => {
+					itemsWidget$().forEach((i) => i.api.expand());
+				},
+				collapseAll: () => {
+					itemsWidget$().forEach((i) => i.api.collapse());
+				},
+				registerItem: (propsConfig?: PropsConfig<AccordionItemProps>) => {
+					const itemProps = accordionItemProps as (keyof AccordionItemProps)[];
+					const config = mergeConfigStores(itemProps, normalizeConfigStores(itemProps, propsConfig?.config), accordionItemConfig);
```
divdavem, line 645: As `accordionItemConfig` does not include the extra props (that can be mentioned in `itemProps`), I am wondering if it actually works to specify a slot at the accordion level that will be used by default by its child accordion items.

quentinderoubaix: Indeed, I have updated this PR with a new commit (easier to review, seeing the size of this PR) that handles this.
I also fixed stackblitz in local, which was broken.

divdavem (approved):

quentinderoubaix:

PR #754, opened 2024-05-02
feat: lockfiles

quentinderoubaix:


demo/scripts/updateStackblitzPkgs.js
```diff
@@ -0,0 +1,22 @@
+import {exec as execSync} from 'child_process';
```
divdavem, line 1:
```suggestion
import {exec as execAsync} from 'child_process';
```

.github/workflows/build.yml
```diff
@@ -85,6 +85,9 @@ jobs:
         run: |
           npm whoami
           node ./scripts/npmPublish.js --provenance
+          npm run -w demo update:stackblitzPkgs ${{ inputs.version }}
+          npm run build -w demo
+          git commit -a -m "feat: update stackblitz packages to ${{ inputs.version }}"
```
divdavem, line 90: You should move the creation of the version tag after this command if you want this new commit to be pushed.

divdavem (approved):

PR #755, opened 2024-05-02
test: stackblitz samples tests

divdavem:


divdavem:
Note that as stackblitz sample tests are much slower than other tests, and as playwright create shards that do not take into account tests execution time, some jobs take 3min while the slowest one takes nearly 18min.
It would be good to have |this issue](https://github.com/microsoft/playwright/issues/17969) solved in playwright.

quentinderoubaix (approved):
LGTM !

divdavem:
@quentinderoubaix Thank you for reviewing and merging my PR!

PR #775, opened 2024-05-21
fix(renovate): remove self-hosted only renovate options from renovate…

quentinderoubaix:
… config

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,5 @@
 {
 	"$schema": "https://docs.renovatebot.com/renovate-schema.json",
-	"repositories": ["AmadeusITGroup/AgnosUI"],
-	"platform": "github",
 	"extends": ["config:recommended", ":dependencyDashboard", "group:allNonMajor"],
 	"ignorePresets": [":prHourlyLimit2", ":prConcurrentLimit10"],
 	"schedule": ["every weekend"],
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -20,17 +18,6 @@
 				"fileFilters": ["**"]
 			}
 		},
-		{
-			"matchPackageNames": ["@angular/core"],
-			"postUpgradeTasks": {
-				"commands": [
```
andreascorti, line 26: Why did you remove angular commands? do they execute by default now?

renovate.json
```diff
@@ -42,7 +29,6 @@
 			"groupSlug": "github-actions"
 		}
 	],
-	"allowedPostUpgradeCommands": ["^npm install$", "^cd angular && npm run ng update .*", "^rm package-lock.json$"],
```
andreascorti, line 45: You removed `"^rm package-lock.json$"` but it is used at @line 16. Does it have any impacts?

andreascorti, line 45: You removed `"^rm package-lock.json$"` but it is used at https://github.com/line 16. Does it have any impacts?

renovate.json
andreascorti, line 26: Do they execute by default now?

andreascorti (commented):
Not sure about the relation between postUpgradeTasks and self-hosted. It is not clear to me how the self-hosted configuration `allowedPostUpgradeCommands` impact `postUpgradeTasks` which are listed in the standard configuration of renovate.

quentinderoubaix:
A new issue was opened to handle this #920

PR #785, opened 2024-05-29
refactor(react): inline useDirective

divdavem:


divdavem:
@quentinderoubaix Thank you for reviewing and merging my PR!

PR #786, opened 2024-05-31
fix: use semver minor versions for peer dependencies

quentinderoubaix:


divdavem (approved):

quentinderoubaix:

Issue #424, opened 2024-01-29
[Doc] Add typedoc


quentinderoubaix:
**Description**
Add typedoc documentation under /api.

PR #787, opened 2024-05-31
feat(accordion): simpler names for accordion items

quentinderoubaix:


core/src/components/accordion/accordion.ts
```diff
@@ -224,7 +166,7 @@ export interface AccordionItemApi {
 	initDone(): void;
 }
 
-export interface AccordionItemDirectives {
+export interface itemDirectives {
```
andreascorti, line 169: aren't always with capital the interfaces?

andreascorti (commented):
I like the idea to make simpler for the end user and remove some redundancies, but it has few drawbacks to consider in my opinion:
1) in the API documentation, now props are scattered between what applies to item or container etc
2) you have an exception for the className, it could have been the case for other props in some other components if we want to uniform this way of writing props without prefixes
3) the code in the core become less readable because of the prefix removal (see for example line 550 in core `accordion.ts`   

I'm ok in general, but maybe to check with others

quentinderoubaix:
> I like the idea to make simpler for the end user and remove some redundancies, but it has few drawbacks to consider in my opinion:
> 
> 1. in the API documentation, now props are scattered between what applies to item or container etc
> 2. you have an exception for the className, it could have been the case for other props in some other components if we want to uniform this way of writing props without prefixes
> 3. the code in the core become less readable because of the prefix removal (see for example line 550 in core `accordion.ts`
> 
> I'm ok in general, but maybe to check with others

thanks for the review @andreascorti !

1. the api documentation was already kind of incomplete (default values not present for the Accordion, no doc for the item itself...) and will be updated with issue #424 
2. indeed, the code of the accordion core has exceptions / complexity that other components do not have..

ValentinNelu (approved):

andreascorti (approved):

PR #801, opened 2024-06-13
feat: generated react bootstrap component types

quentinderoubaix:


.gitignore
```diff
@@ -14,3 +14,4 @@ generated
 vite.config.ts.timestamp-*
 verdaccio/storage
 blob-report
+**/*.gen.ts
```
divdavem, line 17:
```suggestion
*.gen.ts
```

core-bootstrap/src/components/modal/modal.ts
```diff
@@ -48,7 +54,15 @@ interface ModalExtraProps<Data> {
 export interface ModalState<Data> extends CoreState, ModalExtraProps<Data> {}
 export interface ModalProps<Data> extends CoreProps, ModalExtraProps<Data> {}
 
-export type ModalWidget<Data> = Widget<ModalProps<Data>, ModalState<Data>, ModalApi, ModalActions, ModalDirectives>;
+type ModalApiPatch<Data> = {
```
divdavem, line 57: Why did you do this intermediate type?

core-bootstrap/src/components/modal/modal.ts
```diff
@@ -48,7 +54,15 @@ interface ModalExtraProps<Data> {
...
+	/**
+	 * Method to change some modal properties.
+	 */
+	patch: ModalWidget<Data>['patch'];
+};
+export interface ModalApi<Data> extends Omit<CoreApi, 'patch'>, ModalApiPatch<Data> {}
```
divdavem, line 63:
```suggestion
export interface ModalApi<Data> extends Omit<CoreApi, 'patch'> {
	/**
	 * Method to change some modal properties.
	 */
	patch: ModalWidget<Data>['patch'];
}
```

react/headless/src/config.tsx
```diff
@@ -3,14 +3,13 @@ import {createWidgetsConfig, type WidgetsConfigStore, type WidgetsConfig as Core
 import {computed} from '@amadeus-it-group/tansu';
 import type {ReactNode} from 'react';
 import {createContext, useContext, useEffect, useMemo} from 'react';
-import type {AdaptPropsSlots} from './types';
 import {useWidget} from './utils/widget';
 import {usePropsAsStore} from './utils/stores';
 
 export * from '@agnos-ui/core/config';
 
 export type WidgetsConfig = {
```
divdavem, line 11: It looks like this type is now an exact (verbose) copy of `CoreWidgetsConfig`. You could simply remove it, right?

scripts/bootstrapTypes/genTypeExports.ts
```diff
@@ -0,0 +1,147 @@
...
+	// We now resolve all necessary imports to add on top of the generated file.
+	// For this we go through all files in core-bootstrap and core to resolve type-only imports by name / location.
+	// Once this map is built, we change the `@agnos-ui/core` location to be `@agnos-ui/${framework}-headless` instead.
+	const mapImportsByModule = new Map<string, Set<string>>();
+	const visited = new Set();
+	const queue: string[] = [`@agnos-ui/core-bootstrap/components/${component}/${component}`];
+	visited.add(queue[0]);
+	visited.add('@agnos-ui/core/components/commonProps');
+
+	while (queue.length) {
```
divdavem, line 96: Rather than re-implementing an exploration of files to find types as you do here, I would suggest using typescript to resolve types, cf [my commit here](https://github.com/divdavem/AgnosUI/commit/wipGenerateType)
What do you think?

quentinderoubaix: Indeed, thanks for the suggestion !

divdavem (approved):
Thank you for your PR. It is good :+1: 
As discussed, I will open [another PR](https://github.com/AmadeusITGroup/AgnosUI/pull/805) to refactor the part that generate type imports.

quentinderoubaix:
@divdavem thanks for the review !

PR #810, opened 2024-06-18
feat: children instead of slotDefault

quentinderoubaix:


angular/bootstrap/src/components/modal/modal.component.ts
```diff
@@ -298,11 +298,11 @@ export class ModalComponent<Data> extends BaseWidgetDirective<ModalWidget<Data>>
...
 	ngAfterContentChecked(): void {
 		this._widget.patchSlots({
-			slotDefault: this.slotDefaultFromContent?.templateRef,
+			children: this.slotDefaultFromContent?.templateRef,
 			slotFooter: this.slotFooterFromContent?.templateRef,
 			slotHeader: this.slotHeaderFromContent?.templateRef,
 			slotStructure: this.slotStructureFromContent?.templateRef,
 			slotTitle: this.slotTitleFromContent?.templateRef,
-		});
+		} as any);
```
quentinderoubaix, line 306: Workaround due to Typescript behavior when filerting keys by their generic value...
After spending too much time trying to find a cleaner way to do this, am adding this little dirty trick.

Any better suggestion is welcome !

andreascorti (approved):
I did the review on the big PR, and I don't have a strong opinion on the types you created in the PR; I may let @divdavem comment on that.

divdavem (approved):

quentinderoubaix:
thanks for the reviews @divdavem @andreascorti !

PR #814, opened 2024-06-19
Slots modal

quentinderoubaix:


quentinderoubaix:
thanks for the review @andreascorti !
as discussed, we will brainstorm with the team about title / header

PR #818, opened 2024-06-19
Slots select

quentinderoubaix:


svelte/bootstrap/src/components/select/Select.svelte
```diff
@@ -84,13 +86,15 @@
...
+					<Slot slotContent={$itemLabel$} props={{state: $state$, widget, itemContext}} let:component let:props>
+						<svelte:fragment slot="slot" let:props><slot name="itemLabel" {...props} /></svelte:fragment>
 						<svelte:component this={component} {...props}>
 							<svelte:fragment slot="badgeLabel" let:itemContext let:state let:widget
 								><slot name="badgeLabel" {itemContext} {state} {widget} /></svelte:fragment
 							>
-							<svelte:fragment slot="item" let:itemContext let:state let:widget><slot name="item" {itemContext} {state} {widget} /></svelte:fragment>
+							<svelte:fragment slot="itemLabel" let:itemContext let:state let:widget
+								><slot name="itemLabel" {itemContext} {state} {widget} /></svelte:fragment
+							>
```
andreascorti, line 97: is it normal this formatting?

andreascorti (approved):
LGTM

svelte/bootstrap/src/components/select/Select.svelte
quentinderoubaix: don't really like it either... but Svelte 5 removes all that stuff :)

PR #821, opened 2024-06-20
feat: show stackblitz buttons in dev mode only when running dev:stackblitz

quentinderoubaix:


divdavem (approved):

quentinderoubaix:

PR #844, opened 2024-07-09
feat(demo): Show CSS / SCSS variables for a component

### LINKED ISSUE #591: [AgnosUI]: add the documentation of the CSS var in the demo site
- [x] Slider
- [x] Select

andreascorti:
closes #591 
image


demo/src/lib/layout/Style.svelte
```diff
@@ -0,0 +1,11 @@
...
+	import Code from './Code.svelte';
+	export let style: any;
+	let code = '';
+	async function getCode(style: any) {
+		code = await style.css();
+	}
+	$: void getCode(style);
+</script>
+
+<Code {code} language="scss" />
```
quentinderoubaix, line 11: If we want the look to be consistent to the Examples pages, we should wrap it:
```svelte
<div class="mb-4 py-2 px-0 px-sm-3">
   <Code ... />
</div>
```

PR #846, opened 2024-07-12
eslint rule to check and fix default values in tsdoc

### LINKED ISSUE #845: Add a custom linter to force the component props to have the jsdoc @defaultValue properly filled
**Context**
The interfaces Props and State for each component do not share the `defaultValue` for their properties.
This tag is helpful when users check with VSCode documentation of certain properties + when generating type documentation.

**Implementation**
We propose to have a custom linter enforce the writing of those `@defaultValue` tags, based on the functions `get${component}DefaultConfig`.

divdavem:
closes #845

### DIFF
--- a/angular/bootstrap/src/components/accordion/accordion.component.ts
+++ b/angular/bootstrap/src/components/accordion/accordion.component.ts
@@ -194,8 +194,11 @@ export class AccordionItemComponent extends BaseWidgetDirective<AccordionItemWid
 	 * The transition to use for the accordion-item body-container when the accordion-item is toggled.
 	 */
 	@Input('auTransition') transition: TransitionFn | undefined;
+
 	/**
 	 * CSS classes to be applied on the widget main container
+	 *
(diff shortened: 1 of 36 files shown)


quentinderoubaix (commented):
Thanks for the PR !
Looks good to me.

I have noticed however that the command `npm run lint -- --fix` can fail and log an error, even though it actually was able to fix the actual issue...

To reprocude, you can change the default value of `closeOthers` in the `defaultAccordionConfig` and run the command `npm run lint -- --fix`. The command fails for core-bootstrap but fixed the jsdoc in core. Probably because eslint processed core-bootstrap before core, thus the fix of the jsdoc for the AccordionState/Prop was not done yet ?

divdavem:
@quentinderoubaix Thank you for your review! I have added a few changes in `package.json`, as discussed.

quentinderoubaix (approved):

PR #865, opened 2024-07-25
fix: update sass command to fix css source map

quentinderoubaix:


divdavem (approved):

quentinderoubaix:

PR #875, opened 2024-07-31
feat: extract duplicate load functions

quentinderoubaix:


demo/src/routes/docs/[framework]/daisyUI/getMenu.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function getMenu(component: string) {
+	const componentCapitalized = (component.slice(0, 1).toUpperCase() + component.slice(1)) as Capitalize<keyof WidgetsConfig>;
+	const componentNames = Object.keys(daisyUIMetadata);
+	const index = componentNames.indexOf(componentCapitalized);
+	const isFirst = index === 0;
+	const isLast = index === componentNames.length - 1;
+
+	return {
+		...daisyUIMetadata[componentCapitalized],
```
Comment (divdavem) [inline demo/src/routes/docs/[framework]/daisyUI/getMenu.ts:12]: For consistency with the other `getMenu` fonction, maybe it would be better to explicitly write the keys you take from `daisyUIMetadata[componentCapitalized]`.

divdavem (approved):
Thank you for your PR! :+1:
It is good to reduce code duplication!

PR #915, opened 2024-09-10
fix: sass exports

quentinderoubaix:
cascading patch


divdavem:
I am sorry I did not correctly check that in PR #914 that you previously opened.
I think we should always first open PRs to target the main branch. And then if a patch is needed, we should cherry-pick (with -x) from the main branch to the release branch.

PR #916, opened 2024-09-10
fix: patch build on release branches

quentinderoubaix:


divdavem (approved):

.github/workflows/build.yml
```diff
@@ -89,6 +86,7 @@ jobs:
           git tag v${{ inputs.version }}
           git show HEAD
           git push origin v${{ inputs.version }}
+          git push origin release/$(node scripts/versionDocFolder.js "${{ inputs.version }}")
```
divdavem, line 89: This will not work if the branch does not exist.

divdavem: You can do something like this:
```
RELEASE_BRANCH=release/$(node scripts/versionDocFolder.js "${{ inputs.version }}")
git checkout -B $RELEASE_BRANCH HEAD
git push origin v${{ inputs.version }} $RELEASE_BRANCH
```

divdavem (approved):

PR #926, opened 2024-09-18
feat: use @vitest/eslint-plugin instead of eslint-plugin-vitest

quentinderoubaix:


andreascorti (approved):
LGTM @quentinderoubaix !

quentinderoubaix:

PR #963, opened 2024-10-22
Angular signal props and BaseWidgetDirective upgrade

quentinderoubaix:


angular/bootstrap/src/components/progressbar/progressbar.component.ts
```diff
@@ -159,19 +154,22 @@ export class ProgressbarComponent extends BaseWidgetDirective<ProgressbarWidget>
...
+
+	constructor() {
+		super(
+			callWidgetFactory({
+				factory: createProgressbar,
+				widgetName: 'progressbar',
+				defaultConfig,
+				afterInit: (widget) => useDirectiveForHost(widget.directives.ariaDirective),
+				slotTemplates: () => ({
+					children: undefined,
```
andreascorti, line 168: as discussed, we should either filter out children, or (better) remove the possibility to have both a slot and a ng-content

andreascorti (commented):
Thanks @quentinderoubaix ! some comments already discussed with you

angular/bootstrap/src/components/progressbar/progressbar.component.ts
quentinderoubaix: as discussed, it's better to provide a directive for type checking ng-templates

angular/demo/bootstrap/src/app/samples/rating/config.route.ts
```diff
@@ -1,13 +1,13 @@
...
 import {Component, signal} from '@angular/core';
 
 @Component({
 	standalone: true,
-	imports: [RatingComponent],
+	imports: [RatingComponent, RatingStarDirective],
 	providers: [provideWidgetsConfig()],
 	template: `
-		<ng-template #custom let-fill="fill">
+		<ng-template #custom auRatingStar let-fill="fill">
```
andreascorti, line 10: Question: is it mentioned somewhere in the documentation?

andreascorti (approved):
All comments and discussion have been addressed 
LGTM
Thanks @quentinderoubaix !

PR #975, opened 2024-10-25
feat: last runes changes

quentinderoubaix:


andreascorti (approved):
LGTM, leap forward into the way we want to handle toasts :)
Thanks @quentinderoubaix

PR #979, opened 2024-10-28
refactor(pagination): move size property from core to bootstrap-core

divdavem:


quentinderoubaix (approved):

divdavem:

PR #984, opened 2024-10-29
feat: transition directive with no arg

quentinderoubaix:


quentinderoubaix:

PR #988, opened 2024-10-30
docs(headless): add the lifeless keyword in our doc

ExFlo:
### Before submitting the PR, please make sure you do the following:

- [ ] Read the [Contributing Guidelines](https://github.com/AmadeusITGroup/AgnosUI/blob/main/CONTRIBUTING.md)
- [ ] Reference your PR to an issue that was discussed ahead of time
- [ ] The PR should contain a clear description of the problem it solves
- [ ] Ideally, include a test that fails without this PR but passes with it.

### Tests and linting

- [ ] Run the unit tests with `npm run test`
- [ ] Run the linting with `npm run lint`
- [ ] Run the e2e tests with `npm run e2e`

Thank you for contributing to AgnosUI!

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -12,7 +12,7 @@ This is a monorepo organised with multiple npm packages:
 Each framework will include the following packages:
 
 - `headless` delivering the headless wrappers to use the core components with the selected framework
-- `bootstrap` delivering bootstrap-compliant implementations of the components, powered by the headless wrappers and the core components
+- `bootstrap` delivering bootstrap-compliant implementations of the components, powered by the headless/lifeless wrappers and the core components
 - `demo` the demo pages showcasing examples on how to the use the above packages.
 
 Finally, two other npm packages are available for testing purposes:
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 ## Introduction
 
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -13,6 +13,8 @@
 		"alert",
 		"collapse",
 		"components",
+		"headless",
```
quentinderoubaix, line 16: If you want to add the tag `lifeless` to our headless packages, update the script [here](https://github.com/AmadeusITGroup/AgnosUI/blob/main/scripts/buildPackageJson.js#L34)

andreascorti:
> I don't think we should add the `lifeless` word to describe our libraries. `lifeless` literally mean "without life" (or dead?). People may think it is not maintained anymore.
> 
> I know it comes from [this article](https://nerdy.dev/headless-boneless-and-skinless-ui#terms) with the following definitions:
> 
> > Headless: Working components; BYO (bring your own) style.
> > Boneless: Just styles; BYO bones and life.
> > Skinless: Just the markup; BYO skin and life.
> > Lifeless: Just the life; BYO bones and skin.
> 
> I can understand the logic in other definitions: Headless: BYO head Boneless: BYO bones Skinless: BYO skin Lifeless: I would expect: "BYO life", but it is defined in the article as "just the life", literally the opposite of the meaning of the word!

I really don’t like the term either; it has a very bad connotation.

quentinderoubaix:
As discussed, we do not plan to add the `lifeless` keyword in our main documentation.

PR #995, opened 2024-10-30
feat: some slight improvements for home page

quentinderoubaix:


demo/src/routes/+page.svelte
```diff
@@ -13,30 +13,32 @@
 </svelte:head>
 
 <!-- TODO make a list of component and iterate over it when needed -->
-<div class="container-xxl d-flex flex-column align-items-center pt-5">
+<div class="container-xxl d-flex flex-column align-items-center pt-md-5">
```
andreascorti, line 16: just a little remark here: the blur is cut by the header

andreascorti (approved):
Thanks @quentinderoubaix, a small comment, check if you'd like to change or not

demo/src/routes/+page.svelte
quentinderoubaix: indeed, though the vertical space gained was worth this small price

PR #1026, opened 2024-11-20
[Do Not Merge] showcase react bug

quentinderoubaix:
Just a reproduction for issue #1027.
Can be closed once the issue is handled.


quentinderoubaix:
Declined, as the bug has been resolved.

PR #1049, opened 2024-12-05
fix: publishing of base-po and page-objects

quentinderoubaix:


MarkoOleksiyenko (approved):
thanks for the fix, @quentinderoubaix !

PR #1050, opened 2024-12-05
feat: react 19

quentinderoubaix:


divdavem (approved):
Thank you for your PR.
As you said, there is an ugly workaround for React StrictMode in dev, but I don't currently see how to do better, so let's merge the PR as it is for now and we can improve later if we find a better solution.

PR #1071, opened 2024-12-20
fix(slider): Focus handle on keyboard interaction

MarkoOleksiyenko:


MarkoOleksiyenko:
`on:keydown` notation is deprecated in Svelte 5, however, with the new `onkeydown` seems like multiple event handlers rely on event propagation, so with event.stopPropagation() local event handler doesn't work

quentinderoubaix (approved):

PR #1084, opened 2025-01-06
Use $app/state instead of $app/stores

quentinderoubaix:


quentinderoubaix:
Let's wait for clarification from https://github.com/sveltejs/kit/issues/13281

PR #1087, opened 2025-01-08
feat: use clsx in classDirective (and instead of classnames)

divdavem:


react/headless/package.json
```diff
@@ -129,7 +129,7 @@
 	},
 	"dependencies": {
 		"@agnos-ui/core": "0.0.0",
-		"classnames": "^2.5.1"
+		"clsx": "^2.1.1"
```
quentinderoubaix, line 132: can be removed I believe

divdavem:

PR #1116, opened 2025-02-21
fix(slider): Allow drag of the handle initiated on clickable area

MarkoOleksiyenko:


quentinderoubaix (approved):
thx @MarkoOleksiyenko !

PR #1118, opened 2025-02-28
fix(slider): handle focus in range

quentinderoubaix:


MarkoOleksiyenko (approved):
lgtm, thanks @quentinderoubaix !

PR #1125, opened 2025-03-04
fix(demo): use $props.id() rather than hard-coding an id in svelte collapse

divdavem:


quentinderoubaix (approved):

divdavem:

PR #1141, opened 2025-03-19
refactor(stackblitz): install bootstrap from npm instead of using the CDN

divdavem:
This fixes flaky stackblitz e2e tests (caused by the CDN returning an error).


quentinderoubaix (approved):

divdavem:

PR #1148, opened 2025-03-27
feat(carousel): more examples and set status to beta

quentinderoubaix:


angular/demo/bootstrap/src/app/samples/carousel/autoplay-navigation.component.ts
```diff
@@ -0,0 +1,115 @@
...
+		this.progressbar().nativeElement.style.width = '0%';
+		window.requestAnimationFrame(() => {
+			window.setTimeout(() => {
+				this.progressbar().nativeElement.style.transition = `width ${this.api().plugins()!.autoplay.timeUntilNext()}ms linear`;
+				this.progressbar().nativeElement.style.width = '100%';
+			});
+		});
+	}
+
+	ngOnInit() {
```
MarkoOleksiyenko, line 88: as it just done for the effect and and there is a check for the initialized, shouldn't it just be in a constructor? to avoid OnInit hook impl?

angular/demo/bootstrap/src/app/samples/carousel/simpleCustom.route.ts
```diff
@@ -0,0 +1,43 @@
...
+import {Component, ViewEncapsulation} from '@angular/core';
+
+type Photo = {id: string; src: string; className: string; aspectRatio: string; imgClass: string};
+
+@Component({
+	imports: [CarouselComponent, CarouselSlideDirective],
+	template: `
+		<div auCarousel [auSlidesData]="photos" auContainerClass="gap-2" class="custom-carousel" [auSlideClass]="slideClass">
+			<ng-template auCarouselSlide let-src="src" let-aspectRatio="aspectRatio" let-imgClass="imgClass">
+				<img class="w-100" [class]="imgClass" alt="random picsum" [style.aspect-ratio]="aspectRatio" [src]="src" />
```
MarkoOleksiyenko, line 11: just general question for a angular: you think there is no much value of using the ngOptimizedImage?

angular/bootstrap/src/components/carousel/carousel.component.ts
```diff
@@ -68,27 +68,33 @@ export class CarouselSlideDirective<SlideData extends {id: string}> {
...
-						<span class="carousel-control-next-icon"></span>
+					@let nextBtnClassPrefix = state.direction() === 'ltr' ? 'carousel-control-next' : 'carousel-control-prev';
+					<button class="{{ nextBtnClassPrefix }}" [auUse]="directives.scrollNext">
+						<span class="{{ nextBtnClassPrefix }}-icon"></span>
 					</button>
 				}
 			}
 			@if (state.showNavigationIndicators()) {
-				<div [auUse]="directives.tabList">
+				<div class="carousel-indicators" [auUse]="directives.tabList">
```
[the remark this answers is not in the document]
quentinderoubaix: because I would like to use the directive when customizing the `navigation` slot in the autoplay example, but the class `carousel-indicators` was conflicting with my customization...

could happen to a lot of users not satisfied with the current bootstrap navigation look

angular/demo/bootstrap/src/app/samples/carousel/autoplay-navigation.component.ts
quentinderoubaix: sure, will update it

angular/demo/bootstrap/src/app/samples/carousel/simpleCustom.route.ts
quentinderoubaix: for a carousel, you don't really want to load the images with high priority (maybe the first one ? sometimes ?)
so I don't see much value of `ngOptimizedImage` in that example

MarkoOleksiyenko (approved):
lgtm, thanks for the PR @quentinderoubaix

quentinderoubaix:
thx @MarkoOleksiyenko !

PR #1149, opened 2025-03-27
fix(slider): Add overlapping based on the label width

MarkoOleksiyenko:


quentinderoubaix (approved):
lgtm @MarkoOleksiyenko, thx for the fix !

PR #1166, opened 2025-04-07
refactor: simplify callWidgetFactory api

quentinderoubaix:


core/src/types.ts
```diff
@@ -114,7 +115,7 @@ export type WidgetProps<T extends {patch: (arg: any) => void}> = T extends {patc
  * @param props - Optional configuration properties for the widget.
  * @returns An instance of the widget.
  */
-export type WidgetFactory<W extends Widget> = (props?: PropsConfig<WidgetProps<W>>) => W;
+export type WidgetFactory<W extends Widget> = ((props?: PropsConfig<WidgetProps<W>>) => W) & {[FACTORY_WIDGET_NAME]?: string};
```
divdavem, line 118:
```suggestion
export type WidgetFactory<W extends Widget, T extends (props?: PropsConfig<WidgetProps<W>>) => W = (props?: PropsConfig<WidgetProps<W>>) => W> = T & {[FACTORY_WIDGET_NAME]?: string};
```

core/src/types.ts
```diff
@@ -1,4 +1,5 @@
 import type {ReadableSignal, StoreOptions, SubscribableStore, WritableSignal} from '@amadeus-it-group/tansu';
+import type {FACTORY_WIDGET_NAME} from './config';
```
divdavem, line 2: I am not sure if we already had one without this change, but this for sure creates a circular dependency between our files, fortunately only for types, but it would still be better to avoid it.

divdavem (approved):

quentinderoubaix:

PR #1175, opened 2025-05-17
feat: support svelte attach

quentinderoubaix:


fbasso:
Hi @quentinderoubaix , to use attach with directives (action), you may consider using the recent Svelte addition [fromAction](https://svelte.dev/docs/svelte/svelte-attachments#fromAction)

PR #1177, opened 2025-05-30
feat(angular): support angular 20

quentinderoubaix:


tsconfig.json
```diff
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"baseUrl": "./",
 		"target": "es2022",
-		"module": "esnext",
+		"module": "preserve",
```
quentinderoubaix, line 5: Could revert, this is the new default in Angular tsconfig but we don't have to follow this one.

PR #1227, opened 2025-10-22
feat: Add focus directive

MarkoOleksiyenko:


core/src/services/focusElement.ts
```diff
@@ -0,0 +1,62 @@
...
+const processStack = () => {
+	stackProcessPlanned = false;
+	const candidates: HTMLElement[] = [];
+	while (stack.length > 0) {
+		const top = stack[stack.length - 1];
+		if (top.element) {
+			candidates.push(top.element);
+			break;
+		} else {
+			candidates.push(top.previousElement as HTMLElement);
```
divdavem:
```suggestion
		if (top.previousElement) {
			candidates.push(top.previousElement as HTMLElement);
		}
```
Note that `document.activeElement` (which is used to set `previousElement`) can be null under some circumstances as [specified here](https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement#value):

> If the focused element is within a document tree that's not descended from the current document (for example, the focused element is in the main document, and the invoking document is an embedded iframe), then this will be null.

PR #1249, opened 2025-12-03
feat(drawer): mouse move and css units management

fbasso:


angular/demo/bootstrap/src/app/samples/drawer/sizes.route.ts
```diff
@@ -0,0 +1,31 @@
...
+				--bs-drawer-min-size: max-content;
+				--bs-drawer-max-size: calc(100vw - 2rem);
+			}
+		`,
+	],
+	imports: [DrawerComponent, FormsModule, DrawerHeaderDirective],
+	encapsulation: ViewEncapsulation.None,
+})
+export default class SizesDrawerComponent {
+	width = '20rem';
```
MarkoOleksiyenko: not needed anymore

core-bootstrap/src/scss/drawer.scss
```diff
@@ -1,44 +1,85 @@
...
 	flex-direction: column;
 	background: #fff;
 	box-shadow: 2px 0 8px rgba(0, 0, 0, 0.2);
 	z-index: var(--#{$prefix}drawer-z-index);
 
+	// How to use the prefix on ts side ?
+	--#{$prefix}drawer-size: max-content;
+	--#{$prefix}drawer-min-size: 0;
+	--#{$prefix}drawer-max-size: none;
+	--#{$prefix}drawer-splitter-size: 8px;
```
MarkoOleksiyenko: it seems to be replaced by $splitter-size

fbasso: True. Finally, I changed it for the css version, as this is something that we may want to customize.

angular/bootstrap/src/components/drawer/drawer.component.ts
```diff
@@ -64,17 +63,16 @@ export class DrawerBodyDirective {
 	changeDetection: ChangeDetectionStrategy.OnPush,
 	imports: [SlotDirective, DrawerStructureDirective, UseDirective],
```
MarkoOleksiyenko: UseDirective is not needed anymore here

core/src/components/drawer/drawer.ts
```diff
@@ -426,65 +441,103 @@ export const createDrawer: WidgetFactory<DrawerWidget> = createWidgetFactory('dr
...
+				setSize(startSize + direction() * position[isVertical$() ? 'dy' : 'dx']);
 			},
-			dragover: (event: DragEvent) => {
-				event.preventDefault();
+			onMoveEnd() {
+				const isVertical = isVertical$();
+				const drawerElement = drawerElement$()!;
+				const newSize = drawerElement[isVertical ? 'offsetHeight' : 'offsetWidth'];
+				drawerElement.style[isVertical ? 'height' : 'width'] = '';
+				size$.set(newSize);
```
divdavem:
```suggestion
```
Do you really need this line ? Isn't the same thing already done on the last onMove ?

angular/bootstrap/src/components/drawer/drawer.component.ts
```diff
@@ -102,6 +101,9 @@ export const drawerDefaultSlotStructure: SlotContent<DrawerContext> = new Compon
 				<div [auUse]="directives.containerDirective">
 					<ng-template [auSlot]="state.structure()" [auSlotProps]="{state, api, directives}" />
 				</div>
+				@if (state.resizable()) {
+					<div [auUse]="directives.splitterDirective" [tabIndex]="0"></div>
```
MarkoOleksiyenko: if we put a button inside, will this still be focusable?

core/src/components/drawer/drawer.ts
```diff
@@ -149,38 +152,54 @@ export interface DrawerProps extends DrawerCommonPropsAndState {
...
 	 * @defaultValue
 	 * ```ts
 	 * () => {}
 	 * ```
 	 */
-	onHeightChange: (width: number) => void;
+	onVisibleChange: (visible: boolean) => void;
 	/**
-	 * Event to be triggered when the visible property changes.
+	 * Event to be triggered when the minimum size state changes.
```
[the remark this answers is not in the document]
fbasso: It changes in the setSize method :

```ts
isMinimized$.set(size != null && (size < elementSize || size === 0));
```

PR #1254, opened 2025-12-12
feat: Add multiple resize observer directive

MarkoOleksiyenko:


core/src/services/resizeObserver.ts
```diff
@@ -1,45 +1,79 @@
...
+): {
+	dimensionsMap$: ReadableSignal<Map<HTMLElement, ResizeObserverEntry>>;
 	directive: Directive<void, SSRHTMLElement>;
 } => {
-	const {element$, directive} = createBrowserStoreDirective();
+	const {elements$, directive: arrayDirective} = createBrowserStoreArrayDirective();
+	const registeredElements = new Set<SSRHTMLElement>();
+
+	const directive: Directive<void, SSRHTMLElement> = (element) => {
+		if (maxElements !== undefined && registeredElements.size >= maxElements && !registeredElements.has(element)) {
```
divdavem: I think you should not check whether the element is already present. Calling multiple times the directive with the same element can count as multiple usages.

core/src/services/resizeObserver.ts
```diff
@@ -1,45 +1,76 @@
...
+	dimensions$: ReadableSignal<ResizeObserverEntry | undefined>;
+	directive: Directive<void, SSRHTMLElement>;
+} => {
+	const {dimensionsMap$, directive: arrayDirective} = createResizeObserverArray();
+	let firstElement: SSRHTMLElement | null = null;
+
+	const directive: Directive<void, SSRHTMLElement> = (element) => {
+		if (firstElement === null) {
+			firstElement = element;
+		} else if (firstElement !== element) {
```
divdavem:
```suggestion
		} else {
```
I think calling the directive twice should trigger an error, even if it is on the same element.

core/src/services/resizeObserver.ts
```diff
@@ -1,45 +1,76 @@
...
+ * @returns An object containing the store with the dimensions of observed element (ResizeObserverEntry), the directive to be applied to the html element to be observed
+ */
+export const createResizeObserver = (): {
+	dimensions$: ReadableSignal<ResizeObserverEntry | undefined>;
+	directive: Directive<void, SSRHTMLElement>;
+} => {
+	const {dimensionsMap$, directive: arrayDirective} = createResizeObserverArray();
+	let firstElement: HTMLElement | null = null;
+
+	const directive: Directive<void, SSRHTMLElement> = (element) => {
```
divdavem:
```suggestion
	const directive: Directive<void, SSRHTMLElement> = browserDirective((element) => {
```
Then you can remove `as HTMLElement` line 55.

